PR #8, opened 2026-04-08
catalog: refresh listmonk test results

ziadsawalha:
## Summary
- Re-ran test harness for listmonk after promoting to apps/
- Pull time improved (9s → 3s) with cached images
- Startup time slightly faster (10s → 8s)

## Test plan
- [x] `bun run src/test-app.ts listmonk` passes


ziadsawalha:
Content merged into #10 which includes this change plus the larger catalog expansion.

PR #13, opened 2026-04-09
spec: secret encoding properties (base64, hex) via expressions

### LINKED ISSUE #12: spec: secret encoding properties (base64, hex) via expression system
## Problem

Several catalog apps require generated secrets in specific encodings that the spec can't express today:

| App | What it needs | Current workaround |
|-----|--------------|-------------------|
| **Firefly III** (Laravel) | `base64:` prefix + base64-encoded 32 bytes | None — app crashes with HTTP 500 |
| **Monica** (Laravel) | Same `base64:` prefix format | None — app may silently malfunction |
| **Authelia** | Various encoded secret formats | Incomplete Launchfile (in drafts) |

The `generator: secret` strategy produces a hex-encoded random string. There's no way to request a different encoding, and no way to combine a literal prefix with an encoded value.

## Proposed Solution

Extend the **existing expression system** — no new fields, no new syntax. Secrets expose encoding variants as dot-path properties, just like resources expose `$host`, `$port`, `$url`:

```yaml
secrets:
  app-key:
    generator: secret

env:
  # Raw hex (today's behavior, unchanged)
  SESSION_SECRET: "$secrets.app-key"
  
  # Base64-encoded with Laravel's required prefix
  APP_KEY: "base64:${secrets.app-key.base64}"
  
  # Just base64, no prefix
  ENCRYPTION_KEY: "${secrets.app-key.base64}"
```

### Secret property vocabulary

| Expression | Output | Notes |
|-----------|--------|-------|
| `$secrets.key` | `a3f8b2c1d9e7...` | Default (hex), unchanged from today |
| `$secrets.key.hex` | `a3f8b2c1d9e7...` | Explicit hex, same as default |
| `$secrets.key.base64` | `o/iywd6X...` | Base64-encoded (standard, with padding) |

### Why this approach

- **No new syntax** — uses the existing `$secrets.name.property` dot-path resolution
- **Composable with string interpolation** — `"base64:${secrets.key.base64}"` naturally combines literals with encoded values
- **Consistent with resource properties** — resources already expose multiple views (`$url`, `$host`, `$port`); this extends the same pattern to secrets
- **Additive** — existing Launchfiles are unaffected; `$secrets.key` continues to return hex

### Future: key pairs

This pattern naturally extends to key pair generation without additional syntax:

```yaml
secrets:
  signing-key:
    generator: keypair

env:
  JWT_PRIVATE: "$secrets.signing-key.private"
  JWT_PUBLIC: "$secrets.signing-key.public"
  JWT_PUBLIC_B64: "${secrets.signing-key.public.base64}"
```

Key pairs are out of scope for this proposal but the design accommodates them.

## Design Principle Alignment

- **P-1 (Platform-agnostic)**: Encoding is a data format concern, not infrastructure-specific
- **P-6 (No templating)**: This is property resolution, not conditional logic
- **P-7 (Expressions over templates)**: Extends the existing expression vocabulary
- **P-13 (Additive evolution)**: Existing Launchfiles unchanged; new properties are opt-in

## Real-world motivation

At least 3 catalog apps need this today:
1. **Firefly III** — `APP_KEY` must be `base64:` + 32 base64-encoded bytes (Laravel requirement)
2. **Monica** — Same Laravel `APP_KEY` format
3. **Authelia** — Secrets need specific encodings for JWT, session, and storage encryption keys

## Deliverables

- [ ] Design decision entry in DESIGN.md
- [ ] Spec text in SPEC.md (secret property vocabulary table)
- [ ] JSON Schema update (if needed — may not be, since this is expression resolution)
- [ ] SDK resolver update to handle `.base64` and `.hex` on secret references
- [ ] Fix Firefly III and Monica Launchfiles
- [ ] Test coverage in SDK

ziadsawalha:
## Summary

Adds pipe transforms to the expression system so any resolved value can be encoded: `$secrets.key|base64`, `$host|base64`. Composes with string interpolation for literal prefixes: `"base64:${secrets.app-key|base64}"`.

Closes #12

### Design evolution

The initial implementation used dot-path encoding (`$secrets.key.base64`), treating encoding as a property. After researching 15+ expression/template systems (Jinja2, Ansible, Helm, Terraform, Bash, Unix pipes, CSS, GraphQL, JSONPath, XPath, PowerShell, Nix, Kubernetes, Docker Compose), we switched to pipe syntax:

| Candidate | Verdict | Why |
|-----------|---------|-----|
| `$secrets.key.base64` (dot) | Rejected | Ambiguous with property navigation; breaks for non-secret refs |
| `$secrets.key:base64` (colon) | Rejected | Conflicts with `:-` fallback syntax |
| `$secrets.key#base64` (hash) | Rejected | No chainability precedent |
| `base64($secrets.key)` (function) | Rejected | Breaks `$` detection, reads inside-out |
| **`$secrets.key\|base64` (pipe)** | **Chosen** | Universal precedent, unambiguous, chainable, works on any ref |

Full analysis in D-32 in DESIGN.md.

### What changed

**Spec** (SPEC.md):
- New "Pipe transforms" section documenting `|base64` and `|hex`
- Updated expression syntax table with `$ref|base64`
- Noted that dots are for navigation, pipes are for transforms

**Design** (DESIGN.md):
- D-32 with rationale and all rejected alternatives documented

**SDK** (resolver.ts):
- `splitTransforms()` — splits `secrets.key|base64` into path + transforms
- `applyTransforms()` / `applyTransform()` — applies encoding chain after resolution
- Works on ANY reference, not just secrets: `$host|base64`
- Parser updated to recognize `|` in bare refs, braced refs, and template refs
- 6 new tests: hex, base64, Laravel prefix, unknown transform, non-secret ref, parse structure

**Catalog**:
- Firefly III fixed: `APP_KEY: "base64:${secrets.app-key|base64}"` — was HTTP 500, now boots
- Test harness updated to use SDK resolver instead of regex for expression resolution
- Screenshot captured (registration page with pig logo)

### Usage

```yaml
secrets:
  app-key:
    generator: secret

env:
  # Raw hex (unchanged)
  SESSION_SECRET: "$secrets.app-key"
  # Base64 with prefix (Laravel APP_KEY format)
  APP_KEY: "base64:${secrets.app-key|base64}"
  # Works on any reference
  ENCODED_HOST: "$host|base64"
```

## Test plan
- [x] `cd sdk && bun test` — 148 tests pass (6 new)
- [x] `cd sdk && bun run typecheck` — clean
- [x] Firefly III boots with `"base64:${secrets.app-key|base64}"` — no more HTTP 500
- [x] Screenshot captured: registration page renders correctly


sdk/src/resolver.ts
```diff
@@ -279,3 +300,34 @@ function resolvePath(
...
+			// If value looks like hex (even-length, all hex chars), decode as hex bytes first
+			if (/^[0-9a-fA-F]+$/.test(value) && value.length % 2 === 0) {
+				const bytes = new Uint8Array(
+					value.match(/.{1,2}/g)?.map((byte) => parseInt(byte, 16)) ?? [],
+				);
+				return btoa(String.fromCharCode(...bytes));
+			}
+			// Otherwise base64-encode the raw string
+			return btoa(value);
+		}
```
ziadsawalha, line 327: The hex-detection heuristic in `base64` transform is pragmatic but implicit. The spec says `generator: secret` produces a hex string, so this will work correctly for the primary use case. However, two observations:

1. **Spec gap**: SPEC.md says `|base64` produces "Base64-encoded (standard, with padding)" but doesn't document that hex inputs are decoded to bytes first. A non-hex string like `"hello"` would be base64-encoded as the raw string (`aGVsbG8=`), while `"deadbeef"` is decoded as hex bytes first (`3q2+7w==`). This behavior difference should be documented.

2. **Edge case**: A string like `"abcdef12"` matches the hex pattern (even-length, all hex chars) but might be a literal string that happens to look like hex. In practice this is fine because `generator: secret` always produces hex, but the implicit detection could surprise users applying `|base64` to arbitrary string values.

Consider adding a brief note in the SPEC.md pipe transforms section about the hex-aware behavior of `|base64` when applied to secret values. This makes the behavior discoverable rather than surprising.

sdk/src/resolver.ts
```diff
@@ -279,3 +300,34 @@ function resolvePath(
...
+		case "hex":
+			return value;
+		default:
+			return value;
```
ziadsawalha, line 331: Silently returning the raw value for unknown transforms is a reasonable choice aligned with P-6 (parsers ignore unknown fields gracefully). However, this means a typo like `$ref|bsae64` produces no error and no encoding — the value passes through raw. 

This is consistent with how unknown properties resolve to empty string (L-4 in DESIGN.md), but worth noting: when a formal grammar is written for expressions (L-1), the set of known transforms should be part of that grammar so tooling can warn on typos.

catalog/apps/firefly-iii/Launchfile
```diff
@@ -0,0 +1,33 @@
...
+    generator: secret
+env:
+  APP_KEY: "base64:${secrets.app-key|base64}"
+  APP_URL:
+    default: "http://localhost"
+storage:
+  uploads:
+    path: /var/www/html/storage/upload
+    persistent: true
+restart: always
```
ziadsawalha, line 35: Firefly III is missing a `health` check. The app serves HTTP on port 8080 and the metadata confirms `health_check_passed: true`, so the test harness found a working endpoint. Per the catalog review checklist, HTTP-serving components should declare a health check.

Suggested addition:
```yaml
health: /login
```

Also missing the `website` field (`website: https://firefly-iii.org`), though this is minor since `repository` is present.

spec/SPEC.md
```diff
@@ -257,7 +257,7 @@ Generator strategies:
 
 | Generator | Produces | Example |
 |---|---|---|
-| `secret` | Cryptographically random string (suitable for signing keys, tokens) | `a3f8b2c1d9e7...` |
+| `secret` | Cryptographically random hex string (suitable for signing keys, tokens). Use `\|base64` pipe for base64 encoding. | `a3f8b2c1d9e7...` |
```
ziadsawalha, line 260: Good change — explicitly documenting that `secret` produces a *hex* string (not just "random string") is important now that `|base64` exists. The cross-reference to the pipe transform is helpful for discoverability. P-4 (human-writable) benefits from this clarity.

spec/SPEC.md
```diff
@@ -280,6 +280,33 @@ components:
...
+    generator: secret
+
+env:
+  # Raw hex (default)
+  SESSION_SECRET: "$secrets.app-key"
+  # Base64 with Laravel's required prefix
+  APP_KEY: "base64:${secrets.app-key|base64}"
+```
+
+The `|` is unambiguous — dots navigate paths, pipes apply transforms. This distinction matters for future extensibility (e.g., key pair properties like `$secrets.key.private` are navigation, not transforms).
```
ziadsawalha, line 308: This paragraph is well-crafted. The distinction between dots (navigation) and pipes (transforms) is exactly the kind of design clarity that prevents future confusion — especially the forward-looking example of `$secrets.key.private` as navigation. D-32 is well-motivated.

spec/DESIGN.md
```diff
@@ -340,6 +340,17 @@ TOML was considered but rejected: it lacks nested structure depth (tables-of-tab
...
+- Colon (`$secrets.key:base64`) — conflicts with the `:-` default/fallback syntax.
+- Hash/fragment (`$secrets.key#base64`) — no chainability precedent; developers don't associate `#` with transforms.
+- Function syntax (`base64($secrets.key)`) — breaks `$` prefix detection, reads inside-out for chains, requires major parser changes.
+- Field-level encoding (`format: base64` on generators) — doesn't compose with string interpolation for prefixes.
+**Why**: The `|` pipe operator has universal precedent in Unix, Jinja2, Ansible, Helm, and Go templates. It's unambiguous (dots navigate, pipes transform), naturally chainable (`$ref|base64|urlsafe`), has zero YAML conflicts (`|` is only special as a block scalar indicator at value-start), and works on any reference — not just secrets. The parser change is minimal (split on `|` after path parsing). Motivated by Laravel apps (Firefly III, Monica) requiring `base64:`-prefixed keys. See [#12](https://github.com/launchfile/launchfile/issues/12).
```
ziadsawalha: Thorough D-32 entry. Five rejected alternatives with clear reasons is the gold standard for design decision documentation. The cross-reference to issue #12 and the real-app motivation (Firefly III, Monica) makes this easy to evaluate.

One note: the entry mentions `$ref|base64|urlsafe` as an example of chainability, but `urlsafe` is not implemented. This is fine as a future-proofing example, but it might be worth noting in the spec that only `base64` and `hex` are currently defined transforms, with the pipeline being extensible for future additions.

catalog/test/src/launch-to-compose.ts
```diff
@@ -6,6 +6,7 @@
  */
 
 import { stringify } from "yaml";
+import { resolveExpression, isExpression } from "../../../sdk/src/resolver.ts";
```
ziadsawalha, line 9: Good improvement — replacing the ad-hoc regex with the SDK's `resolveExpression` and `isExpression` means the test harness now handles braced expressions, pipe transforms, and fallbacks correctly. This is the right layering: the SDK is the single source of truth for expression resolution (D-10 spirit — one obvious way to do it).

catalog/apps/firefly-iii/Launchfile
```diff
@@ -0,0 +1,33 @@
...
+      DB_HOST: $host
+      DB_PORT: $port
+      DB_DATABASE: $name
+      DB_USERNAME: $user
+      DB_PASSWORD: $password
+secrets:
+  app-key:
+    generator: secret
+env:
+  APP_KEY: "base64:${secrets.app-key|base64}"
```
ziadsawalha, line 27: This is the money line — the whole reason D-32 exists. `"base64:${secrets.app-key|base64}"` composes string interpolation (the `base64:` prefix) with pipe transforms (the `|base64` encoding) cleanly. The fact that this reads naturally and works end-to-end (HTTP 500 → working registration page) is strong validation of the syntax choice.

ziadsawalha (commented):
## Launchfile Steward Review

**Verdict:** ACCEPT (conditional — one checklist item incomplete)
**Confidence:** High
**Change type:** Spec change (D-32 pipe transforms) + Catalog addition (Firefly III) + SDK implementation + CI fixes

### Summary

This PR adds pipe transforms (`$ref|base64`) to the expression system — a well-motivated, cleanly implemented extension that solves a real problem (Laravel's `base64:` APP_KEY requirement, issue #12). The syntax choice follows universal Unix/Jinja2/Helm precedent (P-8), is unambiguous with existing dot-path navigation (P-9), and is fully additive (P-13). The SDK implementation is minimal and correct. The Firefly III catalog entry validates the feature end-to-end. CI fixes are sensible housekeeping.

One checklist item is incomplete: the JSON Schema has not been updated. This is not blocking (the schema doesn't currently validate expression syntax within string values), but it should be tracked.

### 5-Axis Evaluation

| Axis | Result | Notes |
|------|--------|-------|
| Principle alignment | PASS | P-6: just YAML (pipe is not a YAML construct conflict in quoted strings). P-7: simple scaling from `$ref` to `$ref\|base64`. P-8: universal pipe precedent. P-9: dots navigate, pipes transform — one rule. P-13: additive, existing files unchanged. |
| Precedent consistency | PASS | Extends D-2/D-3 expression syntax naturally. D-32 properly records the decision with 5 rejected alternatives. Consistent with existing parser architecture. |
| Catalog impact | ~5-10% | Currently 1/66 apps (Firefly III). BookStack, Strapi, and future Laravel/Rails apps with base64-encoded keys would benefit. Low count but high-value unblock. |
| Complexity cost | Medium | Parser change (not schema-only) — but minimal: `splitTransforms()` is 10 lines, `applyTransform()` is 15 lines. Regex changes are surgical. |
| Reversibility | High | Fully additive. No existing expression changes meaning. Files without `|` parse identically to before. |

### Spec Review Checklist

- [x] SPEC.md fully documents the new behavior (pipe transforms section, expression syntax table)
- [ ] JSON Schema updated — **not done** (though the schema doesn't currently validate expression syntax in string values, this should be tracked)
- [x] Examples provided demonstrating the feature (SPEC.md code block with Laravel pattern)
- [x] DESIGN.md has D-32 entry with thorough rationale and rejected alternatives
- [x] Aligns with the 13 design principles
- [x] Existing valid Launchfiles remain valid

### Catalog Review (Firefly III)

- [x] `name` present and kebab-case
- [x] `image` declared
- [x] `provides` with port, protocol, and `exposed: true`
- [x] `requires` for postgres with `set_env` wiring
- [ ] `health` check — **missing** (see inline comment)
- [ ] `website` field — **missing** (minor)
- [x] `storage` declared
- [x] Uses pipe transforms correctly: `"base64:${secrets.app-key|base64}"`
- [x] `metadata.yaml` with test results present
- [x] Screenshot captured

### CI/Infrastructure

- [x] Build before test ordering fix (SDK needs dist/ for CLI tests)
- [x] Node 22 added (Astro 6 requirement)
- [x] `@astrojs/check` dependency added to both sites
- [x] `wrangler.toml` added with pinned compatibility date
- [x] GitHub Actions use commit SHAs, not floating tags
- [x] `bun-types` → `bun` tsconfig fix

### Inline Comments

See line-level comments below for specific feedback on the SDK implementation and catalog entry.

ziadsawalha:
## Review feedback addressed (56b5fcd)

All three actionable items resolved:

1. **Hex-aware `|base64` behavior documented** — Added paragraph to SPEC.md explaining that hex inputs (as from `generator: secret`) are decoded to bytes first, while non-hex inputs are encoded as raw strings. Makes the implicit behavior explicit.

2. **Currently defined transforms noted** — Both SPEC.md and DESIGN.md now state only `base64` and `hex` are currently defined, with the pipeline extensible for future additions. When the formal grammar lands (L-1), known transforms should be part of it for tooling warnings.

3. **Firefly III health check + website** — Added `health: /login` and `website: https://firefly-iii.org`.

PR #15, opened 2026-04-10
spec: add $app.* expression prefix for platform-injected app properties

ziadsawalha:
## Summary

Adds a `$app.*` prefix to the expression syntax for platform-injected app properties: `$app.url`, `$app.host`, `$app.port`, `$app.name`. The provider's routing strategy resolves these at deploy time, so a Launchfile can reference its own public URL without hardcoding environment-specific values.

```yaml
env:
  PUBLIC_URL: $app.url
  OAUTH_REDIRECT_URI: "${app.url}/oauth/callback"
```

Full rationale and rejected alternatives in [D-33](https://github.com/launchfile/launchfile/blob/main/spec/DESIGN.md#d-33-app-prefix-for-platform-injected-app-properties).

### Motivation — real catalog apps need this today

Five existing catalog apps work around the absence of this feature with either `required: true` (forcing manual config) or hardcoded `localhost` defaults that silently break any non-localhost deploy:

| App | Today | With `$app.*` |
|---|---|---|
| `firefly-iii` | `APP_URL: default: "http://localhost"` (literally hardcoded) | `APP_URL: default: $app.url` |
| `ghost` | `url: required: true` ("Public URL of the Ghost instance") | `url: default: $app.url` |
| `bookstack` | `APP_URL: required: true` | `APP_URL: default: $app.url` |
| `mealie` | `BASE_URL: required: true` | `BASE_URL: default: $app.url` |
| `paperclip` | No `BETTER_AUTH_URL` (auth callbacks broken) | `BETTER_AUTH_URL: default: $app.url` |

The provider always knows the public URL — the spec just needed a way to surface it in expressions. This PR closes that gap for `firefly-iii` directly (commit 3); the other four remain easy follow-ups once the SDK release ships.

### Design

`$app.*` extends the same convention [D-7](https://github.com/launchfile/launchfile/blob/main/spec/DESIGN.md#d-7-resource-properties-as-standard-vocabulary) established for resources — a small standard property vocabulary that providers translate per platform — from "things the app depends on" to "the app itself as a deployment target."

**Standard property set** (the portable vocabulary every provider must support):

| Property | Description |
|---|---|
| `$app.url` | Public URL (e.g. `https://myapp.example.com`) |
| `$app.host` | Public hostname |
| `$app.port` | Allocated public port |
| `$app.name` | App name as deployed |

**Resolution order placement.** `app` is checked **first**, before `secrets` and `components`. The reserved namespace cannot be shadowed by a user-named resource — the same principle that protects `secrets.*` and `components.*`.

**Extensibility.** Providers MAY expose additional `$app.*` properties as platform-specific extensions (`$app.region`, `$app.deployment_id`). Portable Launchfiles should use only the standard set. Unknown properties resolve to empty string, matching [L-4](https://github.com/launchfile/launchfile/blob/main/spec/DESIGN.md#l-4-resource-property-vocabulary-is-implicit).

**Distinguished from `$components.<this>.url`.** Two real differences: (a) `$app.url` is the *public* URL, not the internal component port; (b) it works in single-component mode where there is no component name to reference. They're complementary, not redundant.

### Alternatives considered

| Candidate | Verdict | Why |
|---|---|---|
| `$platform.*` | Rejected | Suggests properties *of the platform* (region, provider name) rather than *of the app as deployed*. Confuses the referent. |
| `$self.*` | Rejected | Ambiguous in multi-component files (which component is "self"?). |
| `$deployment.*` | Rejected | "Deployment" is an orchestrator concept (a specific deploy event). Per P-1 the prefix should describe *the app*. |
| `$components.<this>.url` | Rejected | Requires components to know their own name, doesn't work in single-component mode, gives the *internal* port not the *public* URL. |
| Implicit env-var injection | Rejected | Invisible to AI generators, static analysis, and humans reading the file. Violates P-3 and P-10. |
| Top-level `url:` / `host:` / `port:` fields | Rejected | These would be platform-determined, not author-declared. P-1: the file describes what the app *needs*, not what the platform *will assign*. |
| **`$app.*` reserved prefix** | **Chosen** | Extends D-7's vocabulary pattern, additive per P-13, statically analyzable per P-3, co-located per P-10. |

Full analysis in [D-33](https://github.com/launchfile/launchfile/blob/main/spec/DESIGN.md#d-33-app-prefix-for-platform-injected-app-properties).

### What changed

**Spec** (`spec/SPEC.md`)
- New row in the Expression Syntax table: `$app.prop`
- Resolution order list updated to include `app` as step 1
- New "## App Properties" section with the 4-property table, the routing-strategy explanation, the differences from `$components.<this>.url` and `provides[].port`, and example usage
- One example added to the existing examples block: `PUBLIC_URL: $app.url`

**Design** (`spec/DESIGN.md`)
- D-33 with full rationale and the rejected-alternatives table above
- L-1 updated to reflect the new 6-step resolver order

**Examples** (`spec/examples/single-component.yaml`)
- One new env var demonstrating `PUBLIC_URL: default: $app.url`

**SDK** (`sdk/src/resolver.ts`)
- New optional `app?: Record<string, string | number>` field on `ResolverContext`. Providers populate it from their routing strategy at deploy time.
- New `app`-prefix branch in `resolvePath()`, checked first in resolution order. Unknown properties return undefined so callers can apply `:-` fallbacks.
- Docstring updated to reflect the 5-step resolution order.

**SDK tests** (`sdk/src/__tests__/resolver.test.ts`)
- New `describe` block under REQ-414 with 12 cases:
  - Each standard property: `$app.url`, `$app.host`, `$app.port`, `$app.name`
  - Stringification of numeric `$app.port`
  - Braced form `${app.url}` and composed templates (`${app.url}/oauth/callback`)
  - `:-` fallback when context lacks the property
  - Empty-string behavior when `context.app` is absent
  - Empty-string behavior for unknown properties
  - **Reserved-namespace shadowing check**: a user-named resource called `app` does not shadow `$app.*`
  - Pipe transform composition (`$app.host|base64`) — D-32 + D-33 interaction
  - Cross-talk verification with other context types (`resource`, `secrets`)

**Catalog** (`catalog/apps/firefly-iii/Launchfile`)
- Replaces `APP_URL: default: "http://localhost"` with `APP_URL: default: $app.url`. Validates clean against the SDK CLI.

### Test plan

- [x] `cd sdk && bun run build` — clean
- [x] `cd sdk && bun run test` — 160 pass / 0 fail (148 baseline + 12 new)
- [x] `cd sdk && bun run typecheck` — clean
- [x] Manual: `bun run dist/cli.js validate ../catalog/apps/firefly-iii/Launchfile` — `✓ firefly-iii is valid`

### Principles self-assessment

| Principle | Alignment |
|---|---|
| **P-1** App-focused | The app declares "I need my public URL"; the provider decides what URL to assign. |
| **P-3** Machine-generatable | Typed prefix with a fixed standard vocabulary is statically analyzable; AI generators can produce it from an env var name like `PUBLIC_URL`. |
| **P-5** Provider-translatable | Every provider translates `$app.url` per its own routing strategy: Cloudflare Tunnel, Kubernetes Ingress, lvh.me, ngrok, AWS ALB. |
| **P-6** Just YAML | No new YAML constructs. New prefix token inside existing string values. |
| **P-10** Co-location | Wiring lives next to the env var that consumes it. |
| **P-11** Intent vs. execution | Intent: "use my public URL here." Execution: how the platform assigns one. |
| **P-13** Additive extensibility | New entry in the existing expression vocabulary. Existing Launchfiles unchanged. |

### Backward compatibility

Purely additive. Existing Launchfiles are unaffected. The new `app?` field on `ResolverContext` is optional; existing callers continue to work without it. Providers that have not yet implemented `$app.*` resolution will return empty string for these references — matching the established behavior for unknown resource properties (L-4).

### Open questions

None blocking. Two design points worth a sentence in review:

1. **Extensibility wording.** SPEC.md commits to *"the portable vocabulary every provider must support."* That's stronger than the rest of the spec's MAY/SHOULD tone. Happy to soften to *"portable apps SHOULD rely on the standard set"* if maintainers prefer.
2. **Catalog change in this PR.** Following PR #13's precedent of shipping spec features with at least one real catalog use. If you'd rather land catalog updates as a separate follow-up, the third commit can be cherry-picked out.


ziadsawalha:
**Update:** added a fourth commit (`b1d0faa`) wiring `$app.*` end-to-end through both reference providers — this addresses a real gap in the original PR where the SDK resolver knew about `$app.*` but no provider populated `context.app`, so the `firefly-iii` catalog change would have resolved to empty string for everyone until each provider was updated separately.

**`providers/docker/`** — Adds a `computeAppProperties()` helper that picks the first component with `exposed: true` (in declaration order) and uses its host port (or the declared container port if no override is given) to construct `$app.url = http://localhost:<hostPort>`. The result is injected into `resolverContext.app` at construction time. 5 new tests in `compose-generator.test.ts` cover the main shapes including a roundtrip against the updated `firefly-iii` catalog entry.

**`providers/macos-dev/`** — Same `computeAppProperties()` shape. Also drops the stale local `ResolverContext` interface copy in `env-writer.ts` (the comment said *"since it's not exported"* but `ResolverContext` is in fact exported from `@launchfile/sdk/src/index.ts`) and imports the SDK type directly. Adds an `app` parameter to `buildResolverContext` and updates both call sites in `provider.ts`. 4 new env-writer tests cover property computation, multi-component first-exposed selection, the no-exposed-component fallback (`port: 0`, `url: ""`), and an end-to-end check that `$app.url` in a component's env defaults flows through.

**Multi-component selection rule:** for both providers, `$app.*` always points at the first component (in declaration order) with at least one `exposed: true` provides entry. Apps with multiple exposed components that need a specific one can still use `$components.<name>.url` — `$app.*` deliberately gives a single, predictable answer rather than trying to be clever.

**Test totals:**
- `sdk/`: 160 pass / 0 fail (148 baseline + 12 new for `$app.*`)
- `providers/macos-dev/`: 37 pass / 0 fail (33 baseline + 4 new)
- `providers/docker/`: 28 pass / 0 fail (23 baseline + 5 new)
- Combined: **225 pass / 0 fail**

The original PR description's "What changed" list now also includes the providers section. Happy to fold that into the description directly if you'd prefer that over a comment.

ziadsawalha:
Thanks for the review — both points addressed:

**1. D-33 → L-1 cross-reference.** Fixed in `252efc8`. D-33 now correctly says "6-step order" to match the L-1 update in the same PR. You're right — L-1 was 5 steps before, adding `app.*` at position 1 makes it 6, and the cross-reference was off by one. Pure nit, pure bug, thanks for catching it.

**2. "Must support" wording on the `$app.*` property set.** Keeping it as-is, for the reason you articulated — making `$app.url` / `$app.host` / `$app.port` / `$app.name` *more* portable than the resource vocabulary is genuinely the point of reserving the namespace. If a provider can implement `$components.backend.url` (which every provider can by construction, since the orchestrator assigns the URL), it can implement `$app.url` with the same mechanism. The softer "standard vocabulary" framing in D-7 fits resource types because providers legitimately can't guarantee every resource type is available (no postgres on an iOS deployment target, etc.), but the app always has a URL.

Happy to soften to SHOULD if you change your mind after sitting with it — one-line edit.

Issue #21, opened 2026-04-11
Provider tsconfigs compile __tests__ into dist/, causing vitest to discover tests twice


ziadsawalha:
## Problem

Running `bun run build` followed by `bun run test` inside any of the provider packages (or `sdk/`) causes vitest to discover and execute each test file twice — once from `src/__tests__/*.test.ts` and once from the compiled `dist/__tests__/*.test.js`. Test counts appear doubled even though the same tests are running.

### Reproduction

```
cd providers/macos-dev
rm -rf dist
bun run test         # 47 pass / 0 fail — real count
bun run build
bun run test         # 94 pass / 0 fail — doubled
```

Same pattern in `providers/docker`:

```
cd providers/docker
rm -rf dist
bun run test         # 35 pass / 0 fail
bun run build
bun run test         # 70 pass / 0 fail — doubled
```

## Root cause

`providers/macos-dev/tsconfig.json` and `providers/docker/tsconfig.json` both use `"include": ["src/**/*.ts"]` with `"outDir": "dist"`, which means `tsc` compiles the `__tests__/` directory into `dist/__tests__/`. Vitest's default discovery picks up both the source and the compiled copies.

This is likely visible under `sdk/` as well, though it uses a separate `tsconfig.build.json` — worth checking whether that one excludes tests.

## Impact

- Test counts are misleading (noise, not a correctness regression — the same tests pass twice)
- Tests run 2× longer than necessary
- Potentially masks test ordering or isolation bugs if src and dist versions diverge
- Confuses people reading CI logs

## Proposed fix options

1. **Exclude `__tests__` from the build tsconfigs.** Add `"exclude": ["src/**/__tests__/**"]` to each provider's build tsconfig. Pure — tests don't ship in the published package tarball either (they currently do, as a side effect).
2. **Separate `tsconfig.build.json` per provider** (like `sdk/` already has). Slightly more boilerplate but matches the SDK's convention.
3. **Tighten vitest's include pattern** to only `src/**/*.test.ts`. Simpler but doesn't fix the published-tarball bloat.

Option 1 is probably the minimal correct fix. Option 2 matches the existing SDK convention and is probably the right consistency target.

## Not blocking

This is a build/test hygiene issue, not a correctness regression. Filed as a follow-up during the #17 PR review — surfaced when verifying test totals across the four linked packages.

Issue #20, opened 2026-04-11
Add a Provider Interface doc (spec/PROVIDERS.md) to formalize what new providers must implement


ziadsawalha:
## Problem

The launchfile repo has several excellent spec-side docs — `spec/SPEC.md` (the format), `spec/DESIGN.md` (decision records), `spec/CONTRIBUTING.md` (RFC process) — but there is no document that describes the **provider-side interface**. If someone wants to build a new launchfile provider (a hypothetical Kubernetes provider, Fly.io provider, AWS ECS provider, etc.), there's no single place that tells them what they have to implement or what conventions to follow.

With D-33 (`$app.*`) and D-34 (`commands.bootstrap`) landing, the implicit provider contract has grown enough that this gap is starting to cost real architectural clarity. Specifically:

- When I implemented `dockerBootstrap` in #17 (WIP), I ended up with **three copies** of a `computeAppProperties` helper across `providers/macos-dev/src/env-writer.ts`, `providers/docker/src/compose-generator.ts`, and `providers/docker/src/bootstrap.ts`. All three compute the same thing with trivial variation. The right fix is probably an SDK-level helper like `computeLocalhostAppProperties(launch, hostPorts)` scoped to local-dev providers, but whether to put it in the SDK, in a shared-providers package, or leave it duplicated is really a question about **what the provider interface contract is**, not about the helper itself. Without a provider doc, that question has no canonical answer.
- New providers also need to understand lifecycle command semantics (build before release, release before start, bootstrap after start on user request), the `$app.*` resolution contract, how to wire `ResolverContext`, the capture-extraction pattern, secret generation, state persistence expectations, and the status/logs/down conventions. Today all of that lives in example provider code and commit messages, not in one place.

## Proposal

Add `spec/PROVIDERS.md` (or `spec/PROVIDER-INTERFACE.md`) that documents:

1. **Lifecycle commands** — the canonical stage ordering (`build` → `release` → `start`), well-known vs. named commands, user-invoked stages (`seed`, `test`, `bootstrap`), failure modes per stage, re-run semantics.
2. **`$app.*` contract** — what each of `url`, `host`, `port`, `name` means and where the values come from; the rule that `app` is a reserved namespace checked first in resolution order; the recommendation that local-dev providers use a shared helper (which implies creating one in the SDK as part of this doc's implementation).
[...]

PR #22, opened 2026-04-11
spec+sdk+providers: add commands.bootstrap + nested capture, supersede D-23 placement

### LINKED ISSUE #16: RFC: launch/v1-incubating — a staging zone for under-baked spec proposals, with post_deploy as the first tenant
# RFC: `launch/v1-incubating` — a staging zone for under-baked spec proposals, with `post_deploy` as the first tenant

## TL;DR

Three asks, in order:

1. **Add an incubation mechanism to the spec** — a place for proposed features whose semantics aren't fully figured out yet to live, be implemented by providers, and get real-app validation before graduating to the stable spec. This is a meta-proposal about how the spec evolves, not just about this feature.
2. **Clarify or update D-9 (and P-3) to explicitly allow multi-document YAML.** D-9 lists multi-document YAML alongside custom tags (`!ref`, `!secret`, `!include`), but the rationale given only applies to custom tags. Multi-document `---` is standard YAML 1.2, requires no tag-aware parsers, and is generated correctly by modern LLMs. The parenthetical "for environments" in D-9's rejection list suggests the actual target was environment-variant use, not multi-document in general. I believe the current wording is an inadvertent scope overreach and should be narrowed. Detail and proposed text edits below.
3. **Accept `post_deploy` as the first concrete tenant of the incubation zone.** Post-deploy actions (write config, restart, run admin bootstrap command, capture output) are a real need driven by apps in the catalog, but the semantics are genuinely not figured out yet and I'm not confident enough to propose them for the stable spec. Incubation is the right home.

Fallback path if (2) is rejected: single-document `incubating:` namespace instead of multi-document. Same semantics, less elegant. Fallback path if (1) is rejected: ship `post_deploy` as a plain additive stable field with explicit warning language about design iteration. Neither fallback captures the intent as well as the primary proposal, but both are workable.

## Why incubation, not just "ship it"

Some apps in the catalog need imperative post-start configuration today: write a config file into the running container whose contents depend on `$app.url` (just added in [D-33 / #15](https://github.com/launchfile/launchfile/pull/15)), restart the component so the new config is picked up, run an admin-provisioning CLI command that prints a one-time URL the user needs to see. Paperclip is the concrete case I've implemented downstream and run against real deploys. Gitea, paperless-ngx, docmost, and nocodb look like plausible additional candidates from reading their catalog entries — I haven't personally deployed them with imperative bootstrap, so I'd like the Steward to verify those against the broader catalog context.

The feature isn't controversial. What's controversial is *how confident we are in the semantics* — and I'm not. Honestly:

**I don't know the right answer to most of the design questions.** Here's my current list of things I'd like real-app deploys to answer, not a spec committee:

- **`write-file` template scope**: text only, or also binary via `content_base64`? How does permissions/ownership work (`mode: "0644"`, `owner: user:group`)? Atomic writes or in-place? What happens if the file exists and the template would truncate it — overwrite, fail, or backup?
- **`exec` command format**: string (whitespace-split, matches `commands.*`) or argv list (safer against whitespace-in-arg bugs)? Both? Neither? Does the command go through a shell (`sh -c "..."`) or exec directly?
- **`exec` capture scope**: stdout only (matches D-23 `outputs:`), or also stderr? Some real CLIs emit their most useful output to stderr, and I'm not sure whether that's worth a `source: stdout | stderr | combined` field.
- **Action timeouts**: per-action `timeout:` field, or inherit from `commands.*.timeout`, or no timeout at all (let providers decide)?
- **Idempotency contract**: should the spec *require* actions be idempotent, *recommend* it, or stay silent? What about `write-file` actions that inherently mutate state?
- **Failure semantics**: stop on first failure (what I have), or `continue_on_error:`? What about rollback — is that a provider concern or a spec concern?
- **Cross-action captures**: can a later `exec` reference a value captured by an earlier `exec` (e.g. "invite token from action 1 is used as a flag in action 2")? If yes, what's the syntax? If no, real apps will eventually need this and we'll regret the constraint.
- **Restart semantics**: how long does the provider wait for the component to come back? What if health check fails after restart? Do captured values from earlier actions survive a restart in the same run?
- **State durability**: where do captured values live after the run, how long, and who can read them? This affects whether post-deploy is a one-shot ceremony or a persistent operation log.

These aren't hypothetical edge cases. Every one of them will hit a real app within the first 10 deploys. **I'm not confident enough in any specific set of answers to ask the spec to commit to them as stable.** Committing now means either (a) getting one of these wrong and dealing with a permanent wart on an otherwise-clean spec, or (b) forcing the spec into a deprecation cycle that P-13 was explicitly designed to avoid.

**Incubation is the right shape because the design genuinely needs real-app iteration to converge.** Not because I'm hedging; because I actually don't know yet.

## Why incubation is good for the spec in general

The meta-argument: adding an incubation mechanism pays dividends beyond this one feature. Every future spec proposal will have some version of the same "how confident are we?" problem. A format with P-13 as a Tier 1 inviolable principle (additive extensibility, no syntax changes, no field rename) has a uniquely high cost for getting a field shape wrong, because the obvious fix (rename and deprecate) isn't on the table. Incubation is the pressure-release valve that lets P-13 stay strict at the core while still allowing iteration at the edges.

Looking at state-of-the-art in spec evolution, the pattern of "in-spec incubation with explicit stability markers" is widely practiced and empirically successful:

| System | Mechanism | Stability signal | Graduation |
|---|---|---|---|
| **Kubernetes API versions** | Separate API groups: `batch/v1alpha1` → `batch/v1beta1` → `batch/v1` | Version string contains `alphaN`/`betaN` | New API version; field may rename |
| **Rust feature gates** | `#![feature(name)]` attribute at crate level, nightly compiler only | The attribute itself | Stabilization removes the gate |
| **Python `__future__`** | `from __future__ import name` at module top | Import declaration | Feature becomes default in a later Python version |
| **JSON Schema drafts** | `$schema: draft-07` vs `$schema: 2020-12` | Schema URL encodes draft | Newer draft supersedes |
| **OpenTelemetry** | Per-signal stability marker in spec text: `experimental` / `stable` / `deprecated` | Explicit marker | Stability marker advances over time |
| **Azure ARM API versions** | `2023-03-01-preview` vs `2023-03-01` | `-preview` suffix | Preview version is replaced by stable version |
| **Semver pre-release** | `1.0.0-alpha.3`, `1.0.0-rc.1` | Version string | Strip the pre-release suffix |

Two clear patterns from this survey:

1. **Successful incubation uses explicit version strings, not field-name prefixes.** K8s alpha versions, JSON Schema drafts, Azure preview API versions, semver pre-releases — all of them encode "unstable" in a *version* string, not in a field or resource name prefix. When a proposal graduates, the version string advances; the user-facing names of the fields themselves stay the same.

2. **The `x-` / `X-` prefix pattern has a failed graduation history.** CSS vendor prefixes (`-webkit-`, `-moz-`) and HTTP X-headers both ran into the same wall: once an experimental feature became popular enough to matter, renaming it broke every consumer, and the "experimental" name became permanent. [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648) explicitly deprecates the X-prefix pattern for exactly this reason. This argues against using `x-post-deploy` or similar for incubation: the `x-` convention is fine for OpenAPI-style "this is yours, go wild" vendor extensions, but wrong for "this might become standard" because the graduation path requires a rename.

The pattern that fits launchfile best is the K8s alpha API versioning model, applied to the Launchfile format itself: a second parseable document with its own version string (`launch/v1-incubating`), containing proposals under evaluation. Which brings us to:

## Primary proposal: multi-document YAML with `launch/v1-incubating`

```yaml
version: launch/v1
name: my-app
image: my-app:latest
requires: [postgres]
env:
  APP_URL:
    default: $app.url
provides:
  - protocol: http
    port: 3000
    exposed: true

---
# This document contains incubating proposals — features under evaluation
# that are not yet part of the stable launch/v1 spec. Providers MAY
# implement incubating features; providers that don't understand a proposal
# ignore it. Fields in this document may change or be removed between
# incubating versions. See https://launchfile.dev/incubating/ for the
# current proposal set.
version: launch/v1-incubating

post_deploy:
  - action: write-file
    component: default
    path: /app/config.json
    template: |
      { "publicUrl": "$app.url" }
  - action: restart
    component: default
  - action: exec
    component: default
    command: bootstrap-ceo invite --admin
    capture:
      inviteUrl:
        pattern: "https?://\\S+invite\\S*"
        description: "Admin invite URL"
```

### How it works

**Explicit opt-in.** Users opt in by writing a second document. Files that don't use incubating features are identical to today's files. No change for any existing Launchfile.

**Stability signal.** The `version: launch/v1-incubating` tag is visible at the top of document 2. Anyone reading the file knows immediately that the second document contains unstable content.

**Implementation gradient.** Providers that don't understand `launch/v1-incubating` read document 1, skip document 2, and treat the Launchfile as if the incubating features weren't there. Providers that implement specific incubating proposals advertise which ones (e.g. Launchpad supports `post_deploy@v0.1`) and process document 2 accordingly. An app whose critical functionality depends on an incubating feature is simply incompatible with providers that haven't implemented it — which is the correct failure mode for something explicitly marked unstable.

**Graduation.** When a proposal graduates, its field moves from document 2 to document 1 under the stable `launch/v1` schema. The move is mechanical: the field name stays the same, the semantics stay the same, only the version context changes. Users who want the graduated form can delete their second document and paste the field into document 1. Optionally, graduated fields can be honored in both locations for a deprecation window. Crucially, no field rename is required — this is the lesson from CSS vendor prefixes.

**Removal.** When a proposal is killed (doesn't work, doesn't prove itself), the incubating version bumps (`launch/v1-incubating` becomes `launch/v1-incubating-2` or similar) and the dead proposal is simply absent from the new version. Files still using the killed proposal hit a clear "proposal no longer in incubating" error, not a mysterious unknown-field failure. Users who opted in accepted instability; removal is a valid outcome.

**Structural schema isolation.** The second document has its own schema entirely independent of the stable schema. This matters: if a future incubating proposal needs a structural pattern that doesn't fit the stable schema's shape (e.g. a different approach to `components`), the incubation zone can host it without warping the stable schema. No other shape provides this.

### A stable-only parser gets a complete Launchfile for free

This is the single-biggest engineering advantage of multi-document over every other shape: **a parser that knows only about `launch/v1` and has zero knowledge of incubation reads document 1, gets a complete and valid Launchfile, and runs the app correctly**. No special handling, no "ignore unknown fields" dance, no conditional schema validation. The K8s model again: `kubectl` that doesn't know `batch/v1alpha1` just ignores those resources and proceeds; it doesn't need to be *updated* to correctly-ignore incubating features.

Single-document alternatives (e.g. an `incubating:` namespace) require the stable parser to know about and explicitly skip the namespace. That works via P-13's "parsers ignore unknown fields" rule, but it's operationally different — the stable parser has to *see* the incubating content and decide to ignore it, versus multi-doc where the stable parser simply never reads the second document in the first place.

## D-9 / P-3 scope clarification (the ask embedded in this RFC)

The only thing standing between the proposal above and a clean adoption is D-9's current wording, which appears to reject multi-document YAML. I'd like to argue this wording is an inadvertent scope overreach and should be narrowed.

### The case for unbundling

Reading D-9 verbatim:

> **D-9: Just YAML — no custom tags**
>
> **Decision**: The format uses only standard YAML 1.2 constructs: maps, sequences, scalars.
> **Rejected**: Custom YAML tags (`!ref`, `!secret`, `!include`), multi-document YAML (`---` separators for environments), YAML anchors as a first-class feature.
> **Why**: P-6 and P-3. Custom tags require tag-aware parsers, break generic YAML tooling, and are difficult for AI to generate reliably. Standard YAML parses everywhere and round-trips cleanly.

The rationale sentence — *"Custom tags require tag-aware parsers, break generic YAML tooling, and are difficult for AI to generate reliably"* — is a specific argument against **custom tags**. It doesn't apply to multi-document `---`:

- Multi-document is standard YAML 1.2 syntax, not a custom tag.
- It requires no tag-aware parser — every mainstream YAML parser handles it natively (`yaml.safe_load_all`, `parseAllDocuments`, `Decoder.Decode`, `load_stream`, `serde_yaml::Deserializer`).
- It doesn't break generic YAML tooling — IDE extensions, formatters, and linters all support multi-doc as a standard feature.
- Modern LLMs generate multi-document YAML correctly and routinely (Kubernetes manifests, which are heavily multi-doc, are well within current LLM capability).

So the rationale given in D-9 argues for rejecting custom tags — which is a good decision — but the *list* of rejected things sweeps in multi-document, which the rationale doesn't actually cover. The parenthetical "for environments" on the multi-document item is the clue: the real rejection was of a specific bad use case (multi-document for environment variants, which has a genuine variant-selection problem), but it got written in a way that reads as a general rejection of the syntax.

Reading D-9 most charitably, it bundles three distinct decisions that got lumped into one record:

1. **No custom YAML tags** (`!ref`, `!secret`, `!include`) — correct, well-motivated, keep.
2. **No multi-document `---` for environment variants** — correct, well-motivated (environment variants are a bad pattern regardless of shape), keep but narrow the scope.
3. **Don't build features on top of YAML anchors** — arguably a design discipline ("anchors are an authoring convenience, not part of the spec's semantic model"), not a ban. SPEC.md §YAML Compatibility actually *documents* anchor usage, so this is consistent with treating anchors as allowed-but-not-normative.

The multi-document item in the rejection list is the outlier: the rationale doesn't support a broad rejection, and the narrow case (environment variants) is already well-served by a narrower rejection. I believe the current wording is an unintentional scope overreach.

### P-3 has the same issue

P-3 in DESIGN.md:

> P-3: Machine-generatable. AI can read a repository's structure (Dockerfile, package.json, requirements.txt, docker-compose.yml) and produce a valid Launchfile. The Zod schema provides validation. The format avoids constructs that are hard for language models to produce correctly (custom YAML tags, multi-document streams, complex anchors).

Same bundling pattern. "Custom YAML tags" is a legit concern (rare in training data, easy to malform). "Complex anchors" is arguably legit (anchor-merge-key interactions can confuse LMs). "Multi-document streams" is the one that's empirically outdated in 2026 — LMs handle multi-doc YAML reliably, and the claim in the principle doesn't match current capability.

If D-9 is narrowed, P-3's parenthetical should be narrowed to match.

### Proposed edits

Specific text changes I'd propose if the Steward agrees the scope was overreach:

**D-9 — proposed new wording:**

> **D-9: Just YAML — no custom tags, no DSLs embedded in strings**
>
> **Decision**: The format uses only standard YAML 1.2 constructs: maps, sequences, scalars. Multi-document YAML (`---` separators) is allowed where the two documents have distinct schemas and serve distinct purposes (e.g., stable spec + [`launch/v1-incubating`](...) proposals). Multi-document YAML for environment-variant use (dev/staging/prod as separate documents of the same schema) is rejected — environment overrides are an orchestrator concern (see L-3).
>
> **Rejected**: Custom YAML tags (`!ref`, `!secret`, `!include`); multi-document YAML for environment variants; YAML anchors as a first-class spec feature.
> **Why**: Custom tags require tag-aware parsers, break generic YAML tooling, and are difficult for AI to generate reliably. Standard YAML — including standard multi-document streams — parses everywhere. Environment variants in multi-document form conflate "which document is the spec?" with "which environment am I in?" and are better served by the orchestrator's existing per-environment resolution (L-3).

**P-3 — proposed new wording:**

> **P-3: Machine-generatable.** AI can read a repository's structure (Dockerfile, package.json, requirements.txt, docker-compose.yml) and produce a valid Launchfile. The Zod schema provides validation. The format avoids constructs that are hard for language models to produce correctly (custom YAML tags, complex anchor-and-merge-key patterns).

Both edits are narrowing, not expanding. They preserve the original intent (don't use custom tags, don't build the environment-variant pattern, don't rely on anchors) while allowing multi-document YAML for legitimate structurally-isolated use cases.

### The honest ask

**Was the multi-document listing in D-9 / P-3 an inadvertent scope overreach — meant to reject environment variants but written too broadly — or is there an intentional argument against multi-document for all use cases that I'm missing?** Either answer is useful:

- If it was overreach, please consider the proposed edits above and adopt (or amend) them. This unlocks the primary incubation shape.
- If it was intentional, please explain the reasoning and I'll drop the multi-document proposal and fall back to Shape 2 below.

I'm asking as scope clarification, not as a challenge. Whichever way you answer, the post-deploy feature can still land — the only thing at stake is which incubation encoding it uses.

## Fallback: single-document `incubating:` namespace

If multi-document is intentionally rejected, the single-document incubation shape is:

```yaml
version: launch/v1
name: my-app
# ...stable fields...

incubating:
  version: v1
  post_deploy:
    - action: write-file
      component: default
      path: /app/config.json
      template: |
        { "publicUrl": "$app.url" }
    # ...
```

A reserved top-level `incubating:` namespace whose contents are versioned independently (`incubating.version: v1`, which bumps when the incubation surface itself changes incompatibly). Everything else is identical to the multi-doc proposal: explicit opt-in, visible stability marker, graduation by moving out of the namespace, removal by version bump. No structural schema isolation, but operationally equivalent.

The multi-doc shape is preferred because of the structural isolation (stable-only parsers get a complete Launchfile for free) and because it matches the K8s alpha API precedent more cleanly. The single-doc shape is the fallback if D-9 stays broadly-scoped.

## The first concrete tenant: `post_deploy`

Everything below is what would go *into* the incubation zone, assuming the mechanism is accepted. The shape is tentative — these are starting-point semantics for real-app iteration, not a finished design.

### Problem statement (for this specific tenant)

Some apps need imperative post-start configuration that cannot be expressed by any existing field:

- **Writing a config file into a running container** whose contents depend on values only known at deploy time (`$app.url`, generated secrets, captured output from a previous step).
- **Running an admin-provisioning CLI command** that prints a one-time credential or URL the user needs to see (invite link, reset token, generated admin password).
- **Re-running setup on demand** after the initial deploy — for example when an admin token was lost, or when the app moves to a new URL.

None of this fits into `commands.release`: release runs once, before `start`, so it cannot reference `$app.url` with a routable value; it has no re-run semantics; its failure fails the whole deploy; and its `outputs:` capture (D-23) is per-command, not per-action.

### Real-world motivation

**Primary case: Paperclip** (`catalog/apps/paperclip`, already in the catalog). Paperclip's deploy requires three imperative steps after start: write `/app/config.json` containing the deployment's public URL (only known at deploy time, differs per provider), restart the server so the new config is picked up, run `bootstrap-ceo invite --admin` inside the container which prints an admin invite URL to stdout. This is implemented in a downstream deployment tool (Launchpad, a private Mac-based DevOps agent that consumes `@launchfile/sdk`) and runs successfully against this one app today.

**Suggestions for the Steward to verify against the catalog** (I haven't personally deployed each with imperative bootstrap, but they look plausible from reading their Launchfiles): `gitea` (first admin via `gitea admin user create --admin`), `paperless-ngx` (`createsuperuser` on first deploy), `docmost` (first-time workspace setup), `nocodb` (super-admin creation). The Steward has catalog oversight and is better-positioned to confirm whether any of these actually exhibit the pattern — please treat this as a list of candidates to verify, not claimed verification.

Apps like **Ghost** and **WordPress** intentionally use a first-visit web installer rather than CLI provisioning — those are explicitly *out of scope* for `post_deploy`. This feature is for cases where the bootstrap path is programmatic or extracts a value the user needs to see outside the browser.

**Honest grounding:** one concrete deeply-understood case (Paperclip), a reference implementation that has worked in that one case, and suggestions for the Steward to verify. That is *not* enough to commit the semantics to the stable spec. It is enough to justify putting the feature into the incubation zone so more apps can exercise it.

### Three action types

**`write-file`** — Write content into a component's running container. Supports `$`-expression resolution in `template` (including the just-landed `$app.*` from D-33).

| Field | Type | Required |
|---|---|---|
| `action` | `"write-file"` | yes |
| `component` | `string` | yes (omittable in single-component mode — open question) |
| `path` | `string` | yes |
| `template` | `string` | yes |

**`restart`** — Restart a running component.

| Field | Type | Required |
|---|---|---|
| `action` | `"restart"` | yes |
| `component` | `string` | yes |

**`exec`** — Run a command inside a running container. Optionally capture named values from stdout via regex (mirrors D-23 `outputs:` for consistency).

| Field | Type | Required |
|---|---|---|
| `action` | `"exec"` | yes |
| `component` | `string` | yes |
| `command` | `string` | yes |
| `capture` | `map<string, Capture>` | no |

Each `capture` entry mirrors the D-23 `output` shape:

| Field | Type | Required |
|---|---|---|
| `pattern` | `string` (regex with one capture group) | yes |
| `description` | `string` | no |
| `sensitive` | `boolean` (default `false`) | no |

**Explicit `action:` discriminator field** rather than key-tagged union. Rationale: uniform shape across action types, AI-generator friendly (P-3), easier JSON-schema validation via `oneOf` + `const`, extensible for new action types without breaking existing parsers.

### Execution model (tentative — subject to iteration)

- Actions run **sequentially** in declaration order.
- Execution **stops on first failure**. Captured values from successful prior actions are still surfaced.
- Post-deploy runs **on user request**, not automatically after deploy. This is the distinguishing semantic vs. `commands.release`.
- Post-deploy is **re-runnable**. Actions should be idempotent (spec recommends, does not require — this is one of the open questions).
- Captured values are **persisted and surfaced** by the provider's UI/API after each run.

### Relationship to `commands.release`

The distinction is user-facing and deliberate:

| Aspect | `commands.release` | `post_deploy` (incubating) |
|---|---|---|
| When | After build, **before** start. Runs every deploy. | After start. Runs **on user request**, not automatically. |
| What | A single shell command | An ordered list of typed actions |
| Failure mode | Fails the deploy | Reported to user; does not fail the deploy |
| State | Stateless across deploys | Persisted results; user can re-run |
| Target | The release container (ephemeral) | A running component |
| Output capture | `outputs:` field (component-level, release only) | `capture:` field (per-action) |

**`commands.release` is part of the deploy pipeline; `post_deploy` is a user-invoked post-deploy operation on a running system.** If it ran automatically, it would be `commands.release`.

### Migration path (the shrinking holding area)

An incubation-zone tenant has a natural migration path: every action eventually moves *out* of `post_deploy` as the app's setup matures. Config expressible as env vars → moves into `env:` (e.g. `APP_URL: default: $app.url` — enabled by D-33). Setup that becomes part of the image build → moves into `commands.build`. Bootstrap that becomes web-UI-first → disappears entirely. Actions that become unnecessary → deleted. When the `post_deploy` block is empty → delete it entirely.

This is on top of the incubation-level migration path: if `post_deploy` the *feature* proves itself across many apps, it graduates from the incubation zone into the stable `launch/v1` spec as a regular field. If it doesn't prove itself, it's killed from the incubation zone and apps that used it move to whatever replacement lands.

Two separate graduation paths for two separate concerns: per-app setup simplification (field contents shrink) and cross-ecosystem validation (feature graduates out of incubation). Both are healthy signals.

## Principles self-assessment

| Principle | Assessment |
|---|---|
| **P-1** App-focused | Actions describe what the app needs done (write this config, run this bootstrap), not how the provider runs it. Respects P-1. |
| **P-3** Machine-generatable | Typed discriminated union with `action:` field is schema-validatable. If D-9/P-3 multi-doc scope is updated per the ask above, multi-document YAML is fully within LLM capability as of 2026. |
| **P-5** Provider-translatable | Each action type maps cleanly to provider primitives: `write-file` → `docker cp` / `kubectl exec ... tee` / `fly ssh sftp`; `restart` → `compose restart` / `kubectl rollout restart` / `fly machine restart`; `exec` → `docker compose exec` / `kubectl exec` / `fly ssh console`. Every serious orchestrator can implement all three. |
| **P-6** Just YAML | No custom tags. Standard YAML 1.2 constructs only (including standard multi-document separators, per the proposed D-9 clarification). |
| **P-10** Source of truth co-located | Setup lives in the same file as the declarative spec, not in a sidecar `setup.sh` or README. |
| **P-11** Separate intent from execution | Actions express intent; provider decides execution strategy. |
| **P-13** Additive extensibility | The incubation zone is additive at the format level (new optional second document); post_deploy within it is additive within the incubation schema. Stable `launch/v1` is unchanged. |

**Acknowledged tension with the declarative spirit.** An imperative action list is weakly in tension with "declarative format." Addressed by the incubation framing itself: the feature is explicitly experimental, its semantics are expected to iterate, and the default posture is that it should shrink over time. This isn't a backdoor into permanent imperativeness; it's an escape hatch with both a per-app pressure gradient (fields migrate out) and a spec-level pressure gradient (the incubation zone shrinks as proposals graduate or die).

## Non-goals for v1

- Chained captures between actions (a later `exec` referencing an earlier `exec`'s captured value) — important but should be validated by real apps first
- Binary file writes (`content_base64:`) — add later if a real app needs it
- Conditional execution (`if:` on actions) — definitely not in v1, likely never
- Parallelism or fan-out across actions — sequential only
- Running post-deploy automatically (it is always user-invoked)
- Cross-component orchestration beyond ordered execution
- Rollback semantics (provider concern, not spec)

## Asks, in order

1. **Is the spec open to an incubation mechanism at all?** This is the most important question. If the answer is no — if the preference is "either ship stable or don't ship" — then I'll fall back to proposing `post_deploy` as a plain stable additive field with explicit warning language about iteration, and the incubation discussion becomes moot.

2. **Was D-9 / P-3's multi-document listing intentional or an inadvertent scope overreach?** Please read the unbundling argument above and respond with either "it was intentional, here's why" or "it was overreach, adopt the proposed edits." Either answer is acceptable; only the shape of the incubation encoding depends on it.

3. **Does `post_deploy` make sense as the first concrete incubation tenant?** Assuming (1) is yes, is this a reasonable first case — or would the Steward prefer a different first tenant to establish the incubation pattern, with post-deploy coming later?

Secondary asks:

4. **Graduation/removal policy.** If the incubation mechanism is accepted, what does the policy look like? How long does a proposal incubate before a graduation decision? Who decides? What counts as evidence? I have opinions but would rather hear the Steward's view.

5. **Direction on the post-deploy semantic uncertainty.** Of the ~9 open questions I listed at the top, are any of them "obviously resolved" in the Steward's view, vs. genuinely open? Knowing that would narrow what needs iteration.

## Reference implementation

The downstream implementation (in a private Mac-based DevOps agent called Launchpad that consumes `@launchfile/sdk`) has run the three action types against Paperclip in real deploys. Its current shape uses a second YAML document similar to this RFC's primary proposal, but without an explicit `launch/v1-incubating` version tag. The action semantics, capture mechanism, and execution model are directly transferable to any of the shapes proposed here. I can't link to the source since the repo is private, but I'm happy to share specific file excerpts in the PR discussion if the Steward wants to see concrete implementation choices. If the Steward greenlights incubation and picks a shape, I'll open a PR with:

- `spec:` commit — SPEC.md section for the incubation mechanism, SPEC.md section for `post_deploy` (under the incubating version), schema fragment for the incubating second document / namespace, DESIGN.md D-# records for both the incubation mechanism and `post_deploy` as a tenant, example file
- `sdk:` commit — multi-document parser extension (or namespace parser), incubating schema, test suite under a new REQ-### block
- `providers:` commit — execution implementation for `docker` and `macos-dev`, tests
- `catalog:` commit — Paperclip catalog entry updated to use `post_deploy` in the incubating zone, validated end-to-end

Happy to adjust PR structure or scope per feedback.

---

Thanks for reading. The meta-proposal (incubation mechanism + D-9 scope clarification) is more important than the concrete first tenant. If you're skeptical of one and not the other, please engage with whichever you're willing to — they can land independently, and either one is a step forward.

ziadsawalha:
## Summary

Adds `commands.bootstrap` as a new lifecycle stage for user-invoked post-start setup, introduces a nested `capture:` field on the expanded command form, and removes the top-level `outputs:` field that D-23 originally placed at component level.

The capture mechanism from D-23 (`pattern` / `description` / `sensitive`) is preserved verbatim — only the placement changes, per [D-34](https://github.com/launchfile/launchfile/blob/main/spec/DESIGN.md#d-34)'s P-10 (source of truth co-located) rationale. Capture now lives next to the command it captures from instead of reaching back into `commands:` from a separate top-level block.

Implements Proposal A from the [#16 RFC Steward review](https://github.com/launchfile/launchfile/issues/16#issuecomment-). Closes #16 (Ask 3).

### Example

```yaml
version: launch/v1
name: concentrator
# ... standard fields ...
commands:
  start: "bun run start"
  bootstrap:
    command: "concentrator-cli create-invite --name admin --url $app.url"
    capture:
      invite_link:
        pattern: "https?://\\S+"
        description: "One-time invite link — open to register your passkey"
        sensitive: true
```

Run via `launchfile bootstrap concentrator`. The CLI dispatches to the provider (docker: `docker compose exec`; macos-dev: `spawn`), captures stdout against the regex, masks sensitive values, and exits non-zero on failure.

## Motivation

The primary motivating case is **`catalog/apps/remote-claude-concentrator`**. Its upstream README documents a CLI-by-design first-time setup:

> ```bash
> docker exec concentrator concentrator-cli create-invite --name yourname --url https://concentrator.example.com
> ```
> This prints a one-time invite link. Open it in your browser to register your passkey.

New users can ONLY be created through CLI-generated invite codes — web registration is explicitly disabled. The current catalog Launchfile passes `health_check_passed: true` while being silently unreachable to real users because this step has no declarative home. The Steward's #16 review caught this: *"An app can sit in `catalog/apps/` with `health_check_passed: true` while being silently unreachable to end-users because the Launchfile omits setup steps the upstream README documents."*

Catalog apps in the same shape:

| App | Pattern | Covered by this PR? |
|---|---|---|
| `remote-claude-concentrator` | `concentrator-cli create-invite --url <public>` | ✅ Updated in this PR |
| `paperclip` | `bootstrap-ceo invite --admin` | ✅ Updated in this PR (invite step only; config.json write held for Proposal B) |
| `snipe-it` | `php artisan app:install` (Laravel) | Candidate follow-up |
| Apps exposing CLI `createsuperuser` | Django, Paperless, etc. | Candidate follow-up |

Apps like Ghost and WordPress intentionally use a first-visit web installer rather than CLI provisioning — those are out of scope for `bootstrap`. This feature is for cases where the bootstrap path is programmatic or extracts a value (invite link, password) the user needs to see outside the browser.

### `commands.bootstrap` — a new lifecycle stage

| Aspect | `commands.release` | `commands.bootstrap` |
|---|---|---|
| When | After build, **before** start. Every deploy. | **After** start. **User-invoked**, not automatic. |
| What | A single shell command | Same command-object shape, plus nested `capture:` |
| Failure mode | Fails the deploy | Reported to user; does not fail the deploy |
| State | Stateless across deploys | Re-runnable; captured values persist |
| Target | Ephemeral release container | Running component via provider exec primitive |

Bootstrap exists because `commands.release` can't express *"run this after `start` against the running container and capture what comes out."* Release runs in an ephemeral container, before start, can't reference `$app.url` with a routable value, and fails the whole deploy on error. Bootstrap is the declarative home for the operational knowledge that today lives in READMEs.

### Nested `capture:` — supersedes D-23 placement

D-23 originally placed the capture block as a top-level `outputs:` field at component level. D-34 moves it into `commands.*.capture`, reusing the same `{ pattern, description, sensitive }` shape. The rationale is P-10: capture now lives next to the command it captures from rather than reaching back into `commands:` from a separate block.

### Resolution order and `$app.*`

Bootstrap commands resolve `$`-expressions (including `$app.url` from D-33) at invocation time via the existing SDK resolver. Providers build a `ResolverContext` from persisted state so `$app.url` reflects the actual allocated host port, not whatever was declared at launch time.

## Breaking changes (legitimate under 0.x semver)

1. **Remove `outputs?: Record<string, Output>`** from `Launch`, `Component`, and `NormalizedComponent`, plus the corresponding top-level schemas in `LaunchSchema`, `ComponentSchema`, and `launchfile.schema.json`. **Verified zero usage** in the catalog (`apps/` and `drafts/`, 112 Launchfiles checked) and zero spec examples declaring `outputs:` before removal. Launchfiles that previously used `outputs:` should move the block under the command it captures from — e.g. `outputs.admin_password` with the `release` command becomes `commands.release.capture.admin_password`.
2. **Rename `OutputSchema` → `CaptureEntrySchema`** and `Output` → `CaptureEntry`. The shape is unchanged; only the name reflects its new role as a reusable capture-entry type rather than a component-level "output" vocabulary. Preserved as an export for consumers who reference it from user code.

Both breaks travel together in one changeset (`@launchfile/sdk: minor`) and the linked config propagates the minor bump to `@launchfile/docker`, `@launchfile/macos-dev`, and `launchfile` (the unified CLI).

## What changed

**`spec/`** (2 commits)
- `spec/SPEC.md`: new § Bootstrap stage subsection, new § Command Capture subsection, removed § Outputs (content preserved in § Command Capture, just relocated), components-field enumeration updated
- `spec/DESIGN.md`: new D-34 with full rejected-alternatives table (top-level `outputs:` + `from:`, parallel `capture:` alongside `outputs:`, `outputs:` as deprecated alias, and the original #16 `post_deploy` three-action proposal); header note on D-23 marking the placement supersede (mechanism preserved)
- `spec/schema/launchfile.schema.json`: `outputs` field removed from root and component, new `capture` field on `commandDetail`
- `spec/examples/post-start-capture.yaml`: new example demonstrating the nested shape

**`sdk/`** (2 commits)
- `CommandDetailSchema` gains `capture: z.record(z.string(), CaptureEntrySchema).optional()`
- `OutputSchema` renamed to `CaptureEntrySchema` and moved above `CommandDetailSchema` in the source (eliminates need for `z.lazy()` forward-ref)
- `Output` interface renamed to `CaptureEntry`; `outputs?` fields removed from `Launch` / `Component` / `NormalizedComponent`
- `CommandDetail` and `NormalizedCommand` gain `capture?: Record<string, CaptureEntry>`
- `normalizeCommands()` carries `capture` through; `denormalizeCommands()` keeps the expanded object form whenever `timeout` OR `capture` is set (collapse to string shorthand only for single-field `{ command }`)
- `writer.ts` no longer emits top-level `outputs:`; `reader.ts` no longer passes it through
- 7 new tests under **REQ-415** covering read/write/round-trip of nested capture, the `bootstrap` stage, regex validation via the preserved `CaptureEntrySchema`, and shorthand-collapse behavior for capture-less commands

**`providers/macos-dev/`** (1 commit)
- New `bootstrap.ts` exporting `launchBootstrap({ component?, projectDir? }) → BootstrapResult[]`
- Runs the resolved bootstrap command via `spawn({ shell: false })` with argv split
- Rebuilds the resolver context from persisted state (same pattern as `launchEnv`) so `$app.url` resolves against the running app's actual allocated port
- ANSI-strips stdout before regex matching (critical: CLIs that detect a TTY emit color codes that break simple patterns like `https?://\\S+`)
- Masks sensitive captures as `***` in inline output
- **Does not throw on command failure** — returns `BootstrapResult { ok: false, ... }` and leaves display to the caller, matching SPEC.md § Bootstrap stage's "reported, not deploy-failing" semantic
- 10 unit tests (`bootstrap.test.ts`) covering `extractCaptures` edge cases (single-group, first-group preference, empty result, ANSI stripping, invalid-regex skip, multiple captures, full-match fallback) and `parseDuration` (ms/s/m/h + fallback)

**`providers/docker/`** (1 commit)
- New `bootstrap.ts` exporting `dockerBootstrap({ launch, slug, component? }) → BootstrapResult[]`
- Runs the command via `spawn("docker", ["compose", "-p", project, "exec", "-T", service, ...argv], { shell: false })`
- Same argv-split, ANSI-strip, sensitive-masking, non-throwing semantics as macos-dev
- The CLI (not the provider) parses the Launchfile and passes it in — docker state persists the compose project and host ports but not the source location, so parsing lives in the CLI
- Component-to-service name mapping mirrors `compose-generator.ts` (`default` → `<launch.name>`, named → `<launch.name>-<componentName>`)
- 7 unit tests (`bootstrap.test.ts`) covering docker-specific capture cases

**`packages/launchfile/`** (1 commit)
- New `commands/bootstrap.ts` + new `bootstrap` CLI subcommand in `cli.ts`
- Usage: `launchfile bootstrap [target] [--component <name>]`
- Dispatches to `dockerBootstrap` or `launchBootstrap` based on `deployment.entry.provider`
- Exit code reflects success: 0 if no bootstrap commands ran or all succeeded, 1 if any failed
- Local type shim at `src/macos-dev.d.ts` updated with `launchBootstrap` and `BootstrapResult` (the shim makes `@launchfile/macos-dev` an optional dependency for docker-only installs)

**`catalog/`** (2 commits)
- `remote-claude-concentrator`: adds `commands.bootstrap` for the create-invite CLI flow, plus opportunistic cleanups — `RP_ID: default: $app.host`, `ORIGIN: default: $app.url` (eliminating `required: true`), and `RCLAUDE_SECRET: generator: secret` (which was marked `sensitive: true` with no generator)
- `paperclip`: adds `commands.bootstrap` for the `bootstrap-ceo invite --admin` step plus `BETTER_AUTH_URL: default: $app.url`. The config.json-write and restart pieces of Paperclip's full bootstrap flow are held for Proposal B pending the incubation-mechanism decision on #16.

**`.changeset/`** (2 commits)
- Initial changeset declaring `@launchfile/sdk: minor`
- Expanded to explicitly list all four linked packages (`sdk`, `docker`, `macos-dev`, `launchfile`) at minor so each package's CHANGELOG reflects the entry independently, with a description covering the outputs removal, the `OutputSchema`→`CaptureEntrySchema` rename, the new `bootstrap` stage, provider implementations, and the new CLI subcommand

## Security posture: argv-split, no shell

Both providers run bootstrap commands via `spawn({ shell: false })` with whitespace argv split rather than `shell: true` or `exec()` wrappers. This resolves one of the Steward's #16 open questions (*"string vs argv list"*) in favor of argv, matches Docker CMD / Kubernetes `command`/`args` conventions, and avoids shell injection exposure.

**Trade-off**: shell metacharacters are not supported in `commands.bootstrap` commands. No pipes, redirects, `&&`, or quoted args with spaces. Apps that need shell features should wrap them in an image-level script and invoke that script from `commands.bootstrap`. This is documented in both `bootstrap.ts` source files and in § Bootstrap stage's prose.

## Test plan

Clean test counts (after `rm -rf dist && bun run test` to avoid the pre-existing double-discovery quirk filed as #21):

| Package | Tests | Notes |
|---|---|---|
| `@launchfile/sdk` | 167 pass / 0 fail | 160 baseline + 7 new for D-34 (REQ-415) |
| `@launchfile/macos-dev` | 47 pass / 0 fail | 37 baseline + 10 new bootstrap tests |
| `@launchfile/docker` | 35 pass / 0 fail | 28 baseline + 7 new bootstrap tests |
| `launchfile` (CLI) | 13 pass / 0 fail | 13 baseline; new subcommand exercised by provider unit tests |
| **Combined** | **262 pass / 0 fail** | |

Manual validation:
- [x] `bun run dist/cli.js validate catalog/apps/remote-claude-concentrator/Launchfile` → `✓ concentrator is valid`
- [x] `bun run dist/cli.js validate catalog/apps/paperclip/Launchfile` → `✓ paperclip is valid`
- [x] `bun run dist/cli.js inspect --json ...` on both catalog entries shows `commands.bootstrap` and `capture` surviving normalization intact
- [x] `bun run typecheck` clean across `sdk/`, `providers/docker/`, `providers/macos-dev/`, `packages/launchfile/`

## Principles self-assessment

| Principle | Assessment |
|---|---|
| **P-1** App-focused | `commands.bootstrap` describes what the app needs done (create admin, capture invite), not how the provider runs it (docker exec vs. kubectl exec vs. fly ssh) |
| **P-3** Machine-generatable | Schema-validatable nested shape; AI generators can produce `commands.bootstrap` from a pattern like "this app's README says to run X after install" |
| **P-5** Provider-translatable | Maps cleanly to every orchestrator's "run a command in a running component" primitive — implemented for docker (`compose exec`) and macos-dev (`spawn`); would map to `kubectl exec`, `fly ssh console`, etc. |
| **P-6** Just YAML | Pure schema extension. No new YAML constructs, no custom tags. |
| **P-10** Source of truth co-located | **This is the central win.** Capture now lives next to the command it captures from. Reading `commands.release.capture.admin_password` tells you immediately where the value came from, in a way that `outputs.admin_password` (with the source inferred from prose convention) never could. |
| **P-11** Intent vs. execution | Actions express intent (what needs to run, what to capture); provider decides how to carry them out. Execution-model semantics (sequential, stop-on-failure, re-runnable, reported-not-deploy-failing) are promises about user-visible behavior, not implementation prescriptions. |
| **P-13** Additive extensibility | New optional `capture:` field on an existing expanded-command shape. New `bootstrap` stage name that existing parsers already tolerate via `CommandsSchema`'s open record. Top-level `outputs:` removal is a breaking change but legitimate under 0.x per the evidence above. |

## Follow-ups (filed separately)

- **#20** — [Add a Provider Interface doc (`spec/PROVIDERS.md`)](https://github.com/launchfile/launchfile/issues/20). This PR ends up with three copies of `computeAppProperties` across the two providers (two from PR #15, one new for docker bootstrap). The right fix is an SDK-level `computeLocalhostAppProperties()` helper, but *where* that helper lives is really a question about what the provider interface contract is. Surfacing it as a Provider Interface doc lets the refactor land alongside the broader formalization rather than being answered ad-hoc here.
- **#21** — [Provider tsconfigs compile `__tests__/` into `dist/`, causing vitest to double-discover tests](https://github.com/launchfile/launchfile/issues/21). Pre-existing hygiene issue, not introduced by this PR, surfaced when verifying test totals.
- **Proposal B (incubation mechanism)** — Matomo's `config.ini.php` and Paperclip's `config.json` write steps still have no declarative home. They're explicitly held for the Ask 1 decision on #16 (incubation mechanism for under-baked spec proposals). Paperclip's current catalog update covers only the invite-capture piece; the config-write piece is on the follow-up runway.

## Open questions the Steward specifically listed

- **`exec` command format (string vs argv)** — **resolved in favor of argv** via `spawn({ shell: false })`. String-based command stored in the Launchfile is whitespace-split at invocation time. Safer, consistent with Docker CMD / K8s `command`/`args`.
- **Capture source (stdout vs stderr)** — this PR implements stdout-only, matching D-23. Adding `source: stdout | stderr | combined` is a clean follow-up if real apps need it.
- **Per-action `timeout`** — implemented via the existing `commands.*.timeout` field; `bootstrap` inherits the same shape. Defaults to 120s.
- **Idempotency contract** — spec recommends, does not require. Documented in § Bootstrap stage prose.
- **Cross-action captures** — non-goal for this PR, held for a follow-up if real-app demand materializes.

## RFC trail

- #16 — RFC: `launch/v1-incubating` — a staging zone for under-baked spec proposals, with `post_deploy` as the first tenant
- [#16 Steward updated review](https://github.com/launchfile/launchfile/issues/16) — accepted Proposal A (this PR), deferred Proposal B and the incubation mechanism


packages/launchfile/src/commands/bootstrap.ts
```diff
@@ -0,0 +1,86 @@
...
+	component?: string;
+}
+
+export async function handleBootstrap(
+	target: string | undefined,
+	flags: BootstrapFlags,
+): Promise<void> {
+	const deployment = await resolveDeploymentTarget(target);
+
+	if (deployment.entry.provider === "docker") {
```
ziadsawalha, line 27: Should not live here, but will be addressed in #20

ziadsawalha:
Approving the spec extensions (of course... they are mine!).

Issue #77, opened 2026-06-20
RFC: Execution mode vs. deployment environment — a taxonomy for the spec


ziadsawalha:
# RFC: Execution mode vs. deployment environment — a taxonomy for the spec

## Why this RFC exists (and why it's separate)

Three concepts keep colliding in spec discussions — "dev vs prod," "source vs built," "which environment" — and we've treated them as one. #45 is the symptom: it tried to add dev-mode commands *and* amend L-3 (environment overrides) *and* reserve a namespace, all inside a syntax proposal — three precedent-setting decisions on one feature.

This RFC sets **no syntax and adds no fields.** It establishes the boundary that every command/config/environment proposal after it hangs off, so the follow-ups are mechanical, not philosophical. If ratified, #45 becomes a small follow-up (RFC B, #79).

## The core claim: two orthogonal axes, not one

| Axis | The question | Who knows | What varies |
|------|--------------|-----------|-------------|
| **Execution mode** | *How* is it run — from the working tree, or as a built artifact? | The **app** (only it knows its source entrypoint) | the **commands** |
| **Deployment environment** | *Where / with what config* — local, staging, prod? | The **orchestrator/operator** | the **config** (secrets, URLs, scale, flags) |

They don't collapse into each other:

- **Commands vary by mode, not environment.** Staging and prod run the *identical* start command; they differ only in config. The command changes only when you switch source ↔ artifact.
- **Config varies by environment, not mode.** `DATABASE_URL` differs staging→prod; its meaning is unchanged whether or not an image was built.

So **"dev" is not an environment.** Staging and prod are environments that share a mode (both run the artifact). "dev" is a different *mode* (run from source). Any construct listing `{dev, staging, prod}` as siblings mislabels a binary mode axis with open-ended environment names and carries duplicate values for every artifact environment.

### The litmus test, restated for two axes

The platform-agnostic litmus (P-1) — *a field is app knowledge if changing the deployment target doesn't change its value* — separates app from infra. The corollary that splits the two app-relevant cases:

- Changes with **execution mode** but **not** with deployment target → **app knowledge, in scope.** The source start command is identical on every machine; it varies only by mode. ✅
- Changes with **deployment environment** → **orchestrator knowledge, out of scope.** Per-env secrets, URLs, replicas. ❌ (This is L-3, unchanged.)

## In scope / out of scope

[...]

PR #97, opened 2026-06-28
Start just the parts of an app you want, plus app-state helpers

ziadsawalha:
## What this does

Today `launch up` always starts every part of an app. This lets you start
only the parts you name:

    launch up backend     # starts the backend and the services it needs,
                          # leaves the frontend (and anything else) stopped

- Works on both reference providers — Docker and the macOS dev provider.
- If you name something that isn't a startable part (a typo, or a database
  that belongs to a part rather than being a part itself), it says so clearly
  and stops, instead of guessing.
- The Docker "what's running" summary now lists only the parts you actually
  started — not the whole app.

It also adds three small, self-contained helper functions to the SDK for
tracking what's running:

- one updates a record of the app's state as things happen (a part started,
  became healthy, stopped, got an address),
- one compares two such records and reports what changed,
- one looks up a part's address and returns the right one for whoever's
  asking — a program on your laptop gets `localhost:3001`; a program inside
  the same Docker network gets the internal name.

These are pure functions: they don't touch files or do anything on their own,
they just compute values that tools build on. A separate check warns if two
parts ask for the same database name but with conflicting settings, so a
provider never silently picks one.

## Why

Implements the agreed design decision **D-37** (from #77): you can act on
part of an app, and asking for one part never silently starts others.

## Tested

All green on current `main`: SDK 231 tests, Docker 85, macOS 74, type-checks
clean. The selector was also run for real against Docker — starting only the
chosen part, leaving the rest down, and cleaning up afterward.

4 commits, each scoped to one folder (`sdk:` / `providers:`), per repo convention.

### DIFF
--- a/providers/docker/src/__tests__/summary.test.ts
+++ b/providers/docker/src/__tests__/summary.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, it } from "vitest";
+import { summaryLines } from "../provider.js";
+
+describe("summaryLines", () => {
+	const ports = { frontend: 54000, backend: 54001 };
+
+	it("reports every component when no selector is given", () => {
+		const lines = summaryLines("acme", ports);
+		expect(lines).toHaveLength(2);
+		expect(lines).toContain("  frontend is running at http://localhost:54000");
+		expect(lines).toContain("  backend is running at http://localhost:54001");
+	});
+
+	it("reports only the components actually started under a selector", () => {
+		const lines = summaryLines("acme", ports, new Set(["backend"]));
+		expect(lines).toEqual(["  backend is running at http://localhost:54001"]);
+	});
+
+	it("uses the app name as the label for the default component", () => {
+		const lines = summaryLines("acme", { default: 54000 });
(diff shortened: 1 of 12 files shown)


ziadsawalha:
Thanks — addressed all of it.

- **Closure semantics ratified as D-41**: selecting a component starts its transitive downward `depends_on` closure (+ those targets' `requires`), never reverse-deps/unrelated, idempotent on already-running deps. The satisfy-not-expand interpretation (which contradicted D-16) is recorded as rejected.
- **One shared `selectionClosure()` helper** is now the single source of truth; both providers compute the same start-set (cycle-safe) — the compose-implicit vs hand-narrowed split is gone.
- **macOS** now starts the closure; the `selectComponents` doc-comment and `select.test.ts` are rewritten to the closure rule (+ closure/cycle tests).
- **`resolveRef`** uses `Object.hasOwn` now, so `$components.api.toString` resolves to `""` rather than the inherited method.
- #98 (PROVIDERS.md §5) updated to the same rule for cross-PR consistency.

Green: sdk 237 · docker 85 · macos 74, typecheck/build clean. Ready for re-review.

PR #98, opened 2026-06-28
Add PROVIDERS.md — what a provider actually does with a Launchfile

ziadsawalha:
## What this does

Adds a new document, `spec/PROVIDERS.md`. `SPEC.md` describes the *file*;
this describes what the thing that *runs* your app must do with it — a
"provider" like Docker locally, the macOS dev tool, or a cloud later. Until
now that was scattered across issues and people's heads.

In one place, it explains:

- which commands a provider runs and in what order (install or build, then
  start, plus one-off steps like database migrations);
- running an app from its source code vs. from a pre-built image, and how a
  provider decides per part;
- starting only some parts of an app;
- how a provider builds an app: it must always be able to build from the
  basic info you give (the language plus your commands). A Dockerfile and
  similar files are optional extras a provider may use, but never the *only*
  way to build — so the same app still works on a provider that doesn't use
  Docker;
- how placeholders for URLs, ports, secrets, and storage paths get filled in
  with real values;
- what each existing provider (Docker, macOS) does today;
- what a brand-new provider must do to be compatible.

It also adds a web page so the doc shows up on launchfile.org. (The site
automatically picks up every spec document, but this one had no page to
display it — without the page it would load but appear nowhere.)

## Why

Writes down rules that were only implied before, all based on decisions that
are already approved:

- **D-37** — source vs. image, and acting on parts of an app
- **D-38** — the `install` / `dev` commands and the `source` field
- **D-39** — storage paths filled in by the provider
- **D-40** — the build rule above (build from the basics; Dockerfiles are extras)

Nothing in the file format changes. Existing Launchfiles are unaffected.

4 commits (the document, plus the web page).

### DIFF
--- a/spec/PROVIDERS.md
+++ b/spec/PROVIDERS.md
@@ -1,6 +1,6 @@
 # Launchfile Provider Contract
 
-> **Status: draft for review.** `SPEC.md` defines the *file*; this document defines what a **provider** does with it — the runtime counterpart. Parts marked 📐 depend on RFCs [#77](https://github.com/launchfile/launchfile/issues/77), [#79](https://github.com/launchfile/launchfile/issues/79), [#78](https://github.com/launchfile/launchfile/issues/78) and the cross-invocation state design note, which are not yet ratified. Parts marked ✅ are implemented in the reference providers today.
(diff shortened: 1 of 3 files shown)


ziadsawalha:
Done — applied both inline suggestions verbatim:

- **§10 conformance item 5** now reads "start the selected components' downward `depends_on` closure …" (was satisfy-not-expand).
- **§9 macOS** now reads "narrows `components` to the selected set's downward `depends_on` closure (`selectionClosure`) …".

§5/§9/§10 are now consistent with D-41, and the www-org build is green. Ready for re-review.

PR #100, opened 2026-06-28
fix: restore site search + complete Astro 7 upgrade + guard search in CI

ziadsawalha:
## Summary

Fixes site search on launchfile.dev (broken for ~6 days), completes the Astro 6→7 upgrade across both sites, and closes the CI gap that let the regression ship.

## Root cause

After the dependabot **astro 6.4.8→7.0.0** bump on www-dev (`0b0a860`), launchfile.dev search opened but every query showed *"Search is not available right now."* The index was fine (Pagefind served 200s, direct queries returned results) — the **bundled** `loadPagefind()` threw `ReferenceError: __VITE_PRELOAD__ is not defined`. Under Vite 7, the vite-ignored dynamic `import(url)` is rewritten into a `__vitePreload(…, __VITE_PRELOAD__)` call that leaks an undefined `__VITE_PRELOAD__` into the inline script. A silent `catch {}` hid it. www-org (still on Astro 6) was unaffected — same shared component, different Astro major.

## Changes (one commit per directory)

- **`www-shared:`** — load Pagefind through a `new Function("u","return import(u)")` indirection so the bundler can't rewrite the import (no `__VITE_PRELOAD__`). The URL is a constant path on `location.origin` (no user input) and the sites ship no CSP, so the indirection is safe. Also replaced the silent `catch {}` with a logged one — the silent swallow is *why this shipped unnoticed*.
- **`www-org:`** — upgrade to astro 7.0.0 / @astrojs/cloudflare 14.0.0 (match www-dev). Two Astro 7 breaking changes handled: (1) Astro 7 drops `@astrojs/markdown-remark` from defaults — re-added so the spec/design remark pipeline (remark-breaks + the custom remarkRewriteLinks + shiki) keeps rendering identically; (2) the mermaid CDN import in DocsLayout hit the same `__VITE_PRELOAD__` bug — same indirection fix, plus a guard so mermaid is only fetched on pages with diagrams.
- **`smoke-tests:`** — parameterize the search suite (`registerSearchTests(name, baseURL, term)`) so it runs against the live sites (post-deploy, unchanged default) **or** a locally-served build (pre-merge CI). Adds `serve-build.mjs` (static server with correct content-types). Replaced the content-fragile "no results for gibberish" assertion (Pagefind's loose matching returns a spurious hit for some strings on the larger org index) with a content-agnostic check that targets the real regression: search must *execute* — never the "unavailable" error, past the prompt, to a definite response.
- **`chore:` (CI)** — the `websites` matrix only built the sites; the only functional search test ran **post-deploy against live URLs**, which is why this sat red for ~6 days. Now each built site is served and search-tested **before** deploy, so a regression fails the PR.

## Verification (live, not just green tests)

On clean Astro-7 builds of both sites:
- `__VITE_PRELOAD__` gone from output; typecheck 0 errors; `/design/` renders the full DESIGN.md (103KB); mermaid flowchart renders to SVG on `/spec/expression-syntax/`.
- The parameterized search suite passes **6/6 on both www-dev and www-org** via the exact CI flow (build → serve-build.mjs → playwright): known term → results, search executes without error, opens/closes, click navigates. Zero console errors.

## Scope note

Only the two sites in this repo are bumped. `launchfile-io` (separate repo) is intentionally untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,8 +54,13 @@ jobs:
     name: "Website (${{ matrix.site }})"
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        site: [www-dev, www-org]
+        include:
+          - site: www-dev
+            term: validate
+          - site: www-org
+            term: components
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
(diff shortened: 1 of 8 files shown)


ziadsawalha:
Addressed the steward's non-blocking notes:
- **Note 1** (path-traversal guard) — `dbb33f6`: now uses `path.relative()`/`isAbsolute()` (sibling-prefix safe). Verified a `../../etc/passwd` request 404s.
- **Note 2** (env vs inline interpolation) — `ce4ac87`: matrix values pass through an `env:` block.
- **Note 3** (CSP coupling) — `b112ed5`, `c612e1e`: comments at both `new Function` sites.
- **Note 4** (mermaid not exercised by CI) — tracked as a follow-up (the diagram render is verified manually here but not yet guarded by CI).

Re-verified search 6/6 against both built sites after the changes.

PR #104, opened 2026-06-28
catalog: resolve set_env via the SDK resolver in the test harness

ziadsawalha:
## Summary

The catalog compose-generation test harness resolved `env:` defaults through the SDK resolver, but `set_env` used a separate `resolvePropRef()` stub that only saw the enclosing resource's properties — so `$secrets.*`, `$storage.<name>.path`, and `$app.*` in a `set_env` block resolved to the **raw `"$ref"` string** (or `""`), unlike the real providers. This is a harness-fidelity gap flagged in the #94 (`$storage.*`) review.

## Change

- Route `set_env` through `resolveExpression()` with a proper scoped context — the enclosing resource (single-segment `$url`/`$host`) + accumulated named resources + `secrets` + `storage` + a best-effort `$app.*` — matching the docker / macos-dev providers.
- Unify the `env:`-default path onto the same component context, so `$app.*` resolves there too (it previously only had `secrets`/`storage`).
- Remove the `resolvePropRef()` stub.

The harness assigns **ephemeral host ports** at generation time, so `$app.url`/`$app.port` aren't knowable here — they degrade to `""` via the resolver, exactly as on a real provider that hasn't computed them.

## Verification

- `set_env`: `$url` → postgres URL, `$secrets.api-key` → the generated secret, `$app.host` → `localhost`.
- `env:`: `$storage.data.path` → `/data`, `$app.host` → `localhost`.
- `launch-to-compose.ts` typecheck: no new errors.
- A real catalog app (activepieces) generates compose with **no unresolved `$` refs**.

### DIFF
--- a/catalog/test/src/launch-to-compose.ts
+++ b/catalog/test/src/launch-to-compose.ts
@@ -237,6 +237,10 @@ export function launchToCompose(launch: NormalizedLaunch): ComposeResult {
     }
   }
 
+  // App-wide resource properties for $<resource>.prop set_env resolution, shared
+  // across components like the Docker provider's resourceMap (compose-generator).
+  const resources: Record<string, Record<string, string>> = {};
+
   for (const [componentName, component] of Object.entries(launch.components)) {
     const serviceName =
       componentName === "default" ? launch.name : `${launch.name}-${componentName}`;
@@ -333,7 +337,6 @@ export function launchToCompose(launch: NormalizedLaunch): ComposeResult {
     const dependsOn: Record<string, { condition: string }> = {};
 
     if (component.requires?.length) {
-      const resources: Record<string, Record<string, string>> = {};
       for (const req of component.requires) {
         const backingResult = addBackingService(
           launch.name,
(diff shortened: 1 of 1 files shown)


ziadsawalha:
Addressed steward note **#1** in `6c6441d`: hoisted `resources` to app-wide function scope (exact Docker-provider parity). Verified a cross-resource `$postgres.url` from a sibling resource's `set_env` now resolves.

Tracking the other non-blocking notes as follow-ups: **#2** (`$components.*` still unpopulated in the harness — pre-existing) and **#4** (a characterization test in `catalog/test/`). Note **#3** (`$app.url` is port-less in the harness) is an intentional, documented limitation.

ziadsawalha:
CodeQL caught a real **high-severity `js/insecure-randomness`** that the manual review's CWE pass missed: `generatePort()` used `Math.random()`, and since a secret can declare `generator: port`, routing set_env through the resolver made that value reach a credential context. Fixed in `9c625c4` — `generatePort` now uses `crypto.getRandomValues` (matching `generateSecret`/`generateUuid`). **0 open alerts; CodeQL green.** Port output unchanged in range; resolution re-verified.

PR #157, opened 2026-08-22
Potential fixes for 2 code quality findings

ziadsawalha:
_This PR applies 2/3 suggestions from code quality [AI findings](https://github.com/launchfile/launchfile/security/quality/ai-findings). 1 suggestion was skipped to avoid creating conflicts._

### DIFF
--- a/www-dev/src/pages/learn/specialized-patterns.astro
+++ b/www-dev/src/pages/learn/specialized-patterns.astro
@@ -131,7 +131,7 @@ commands:
     { line: "host:", text: "Declares exactly what host-level access this app requires." },
     { line: "docker: required", text: "Needs the real Docker daemon — Docker-in-Docker won't work." },
     { line: "network: host", text: "Shares the host network to manage container ports directly." },
-    { line: "filesystem: read-write", text: "Persistent state stored on the host filesystem." },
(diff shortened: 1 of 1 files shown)


ziadsawalha:
Closing this one — both suggestions would make the page less correct, not more.

### `docker: required` → `docker: true` is invalid syntax

The SDK schema defines the field as an enum:

```ts
// sdk/src/schema.ts:178
docker: z.enum(["required", "optional"]).optional(),
```

Tested against the real parser:

```
docker: required  -> VALID
docker: optional  -> VALID
docker: true      -> REJECTED
```

The page renders `spec/examples/host-orchestrator.yaml`, which correctly says `docker: required`. This change would print `docker: true` as the annotation label directly above a code block showing `docker: required` — teaching a value the parser rejects.

### `filesystem:` → `host.filesystem:` is inconsistent

Not wrong, but it dots one of four labels while `host:` and `network: host` stay bare. The annotation labels quote lines as they appear in the code block above them; the dotted form is what the surrounding prose uses. Mixing the two in one list is worse than either convention applied consistently.

### Why CI stayed green

`RealWorldExample.astro` renders `line` as a `<code>` label inside a `<dl>` — it is never matched against the YAML. So neither change breaks the build; they only make the docs wrong. Worth knowing for future findings on this component.

No action needed on the underlying findings — the file is correct as it stands.

PR #159, opened 2026-08-22
Potential fix for 1 code quality finding

ziadsawalha:
_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/launchfile/launchfile/security/quality/ai-findings)._

### DIFF
--- a/catalog/apps/gitea/Launchfile
+++ b/catalog/apps/gitea/Launchfile
@@ -16,7 +16,7 @@ provides:
 requires:
   - type: postgres
     set_env:
-      GITEA__database__DB_TYPE: $host
+      GITEA__database__DB_TYPE: postgres
       GITEA__database__HOST: $host
       GITEA__database__NAME: $name
       GITEA__database__USER: $user


ziadsawalha:
Swept the rest of the catalog for this bug shape — this is the only instance.

**Exact shape (a `set_env` key that also has an `env` default):** 113 Launchfiles across `catalog/apps` and `catalog/drafts`, checking top-level and per-component scopes.

```
scanned 113 Launchfiles
gitea   GITEA__database__DB_TYPE   env.default="postgres"  set_env="$host"  <<< MISMATCH
```

One hit, this one.

**Broader shape (a `$expr` that disagrees with what the key name implies):** 173 `set_env` expressions, 149 with a recognisable key shape (`*_HOST`, `*_PORT`, `*_USER`, `*_PASSWORD`, `*_NAME`, `*_URL`). One flag, and it is a false positive:

```
photoprism   PHOTOPRISM_DATABASE_SERVER = "$host:$port"   (name implies $host)
```

That is a deliberate composite. `catalog/drafts/photoprism/Launchfile:17`.

Incidentally photoprism also settles the open question on this PR — it puts its literal driver in `set_env` alongside the `$` values:

```yaml
set_env:
  PHOTOPRISM_DATABASE_DRIVER: "mysql"
  PHOTOPRISM_DATABASE_SERVER: "$host:$port"
```

So `DB_TYPE: postgres` in `set_env` matches existing precedent. Disregard my earlier suggestion about deleting the line instead — this PR is right as it stands.

PR #161, opened 2026-08-22
Potential fixes for 4 code quality findings

ziadsawalha:
_This PR applies 4/5 suggestions from code quality [AI findings](https://github.com/launchfile/launchfile/security/quality/ai-findings). 1 suggestion was skipped to avoid creating conflicts._

### DIFF
--- a/providers/macos-dev/src/resources/postgres.ts
+++ b/providers/macos-dev/src/resources/postgres.ts
@@ -46,25 +46,20 @@ export class PostgresProvisioner implements ResourceProvisioner {
 		}
 
 		// Wait for ready
-		const ready = await shellOk("pg_isready --timeout=10");
-		if (!ready) {
-			throw new Error("PostgreSQL did not become ready within timeout");
-		}
+		await shell("pg_isready --timeout=10", { allowFailure: true });
 
 		// Determine database and user names
 		const resourceName = req.name ?? req.type;
 		const safeName = opts.appName.replace(/-/g, "_");
 		const dbName = existingState?.dbName ?? `launchfile_${safeName}`;
 		const user = existingState?.user ?? `launchfile_${safeName}`;
+		// Security: both are interpolated into the psql commands below. A schema-validated
+		// app name can never fail this, but existingState comes from the on-disk state file,
+		// which is JSON.parse'd without validation (state.ts).
 		if (!SAFE_IDENTIFIER.test(dbName)) throw new Error("Invalid database name");
 		if (!SAFE_IDENTIFIER.test(user)) throw new Error("Invalid database user");
 		const password = existingState?.password ?? generatePassword();
 		const port = DEFAULT_PORT;
-
-		// Security: validate identifiers before using them in shell commands.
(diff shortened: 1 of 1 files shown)


ziadsawalha:
Reviewed. The first hunk is worth keeping; the other three should come out. CI is green (typecheck clean, 89/89 tests pass with this branch applied), so none of this is caught automatically.

### Keep: the identifier checks at lines 59-60

These earn their place, though not for the reason the finding gives. A schema-validated app name can never fail `SAFE_IDENTIFIER` — `NameSchema` is `^[a-z][a-z0-9-]*$` (`sdk/src/schema.ts:11-12`), and `safeName` only turns `-` into `_`, so the generated `launchfile_${safeName}` always matches.

The real exposure is the other branch. `dbName` and `user` fall back to `existingState`, which comes from `JSON.parse(raw) as LaunchState` (`providers/macos-dev/src/state.ts:82`) — a bare type assertion with no runtime validation — and both values are interpolated into `psql` command strings below. Validating them is correct.

Worth a comment saying so, otherwise the next reader deletes the check as redundant.

### Remove: the duplicate validation at lines 64-67

```ts
if (!SAFE_IDENTIFIER.test(dbName)) throw new Error("Invalid database name");   // 59
if (!SAFE_IDENTIFIER.test(user)) throw new Error("Invalid database user");     // 60
const password = ...; const port = ...;                                       // 61-62

// Security: validate identifiers before using them in shell commands.
if (!SAFE_IDENTIFIER.test(user) || !SAFE_IDENTIFIER.test(dbName)) {            // 64-67
    throw new Error("Invalid PostgreSQL identifier for user or database name");
}
```

Same two predicates, and nothing between lines 60 and 64 reassigns `user` or `dbName` — 61 and 62 only declare `password` and `port`. This block can never throw. Dead code.

### Remove: the in-loop check at lines 98-101

Same problem, inside the extensions loop: `dbName` was validated at line 59 and never reassigned, so the re-check can't fire — and it re-tests a loop invariant on every iteration.

### Revert: the `pg_isready` change

This is a behaviour change, not a code-quality fix, and it isn't mentioned in the PR description:

```ts
- await shell("pg_isready --timeout=10", { allowFailure: true });
+ const ready = await shellOk("pg_isready --timeout=10");
+ if (!ready) throw new Error("PostgreSQL did not become ready within timeout");
```

`pg_isready --timeout=10` blocks up to 10s waiting for readiness, so the original reads as a deliberate "wait, then proceed regardless". Turning it into a hard failure means provisioning now aborts on any machine where Postgres needs more than 10s — plausible on a cold `brew services start`.

It may well be the better behaviour, but it needs its own PR and a test. Nothing currently exercises this path: no test in `providers/macos-dev/src/__tests__/` references `PostgresProvisioner` or `pg_isready`, and the suite has no shell-mocking setup to hang one on.

---

I stopped short of pushing these edits myself — stripping `SAFE_IDENTIFIER` checks from a security-sensitive path is exactly the kind of change that should be made deliberately by a human rather than applied by an agent, even when the removal is provably safe.

Issue #120, opened 2026-06-29
RFC: deprecate host.docker → container_runtime (first application of legible evolution)


ziadsawalha:
## RFC: deprecate `host.docker` → `host: { container_runtime: … }` (first application of legible evolution)

The concrete deprecation that motivated the whole thread. It **applies** the legible-evolution principle (#117) to retire the P-1-leaking `host.docker` key in favor of the ratified `container_runtime` capability (#113, syntax in #118), without ever breaking an existing file.

> **Depends on #117** (the legible-evolution principle + deprecation metadata must exist first) **and #118** (`container_runtime` must exist as the replacement). Filed now to make the sequence explicit; it lands after both.

### What changes

- Mark `host.docker` **deprecated**, `replacement: host.container_runtime`, with a migration hint.
- The SDK **normalizes** legacy `host: { docker: required }` → `host: { container_runtime: docker }` (and `optional` → a `supports` entry) at parse time, so existing files validate and run unchanged (P-13).
- Tooling (per #117) reports it: `launchfile lint` flags `host.docker` as deprecated and offers the migration; `launchfile migrate` applies it.
- **Removal** of the legacy key happens only at the next major version, per #117's lifecycle — never before, never silently.

### Why a key, not the block

Because #113 keeps `host:` as the conspicuous home for privileged capabilities, this deprecates only the **`host.docker` key** (and folds `network`/`filesystem`/`privileged` into the same entry form per #118) — a one-key rename with a normalizing alias, the lightest possible first use of the deprecation policy. The block stays.

### Why now (motivation)

`host.docker` is the spec's clearest P-1 leak (it names a product at the intent layer) and its only field with zero adoption *and* an active replacement. It's the ideal first test of #117: a real, low-risk, fully-migratable deprecation.

### Scope

Just the `host.docker` deprecation + normalization + the metadata that drives the tooling. The replacement syntax is #118; the policy/tooling is #117. Nothing here is novel precedent — it's the application.

### Verification (for the implementing PR)

`cd sdk && bun test` proving a legacy `host: { docker: required }` file normalizes to `host: { container_runtime: docker }` with unchanged meaning; the schema validates both legacy and new forms; `launchfile lint` surfaces the deprecation with its migration.

Issue #113, opened 2026-06-28
RFC: ratify the host-capability model, and mark the current host: block provisional


ziadsawalha:
## RFC: ratify the host-capability model, and mark the current `host:` block provisional

This is the **boundary/ratification** RFC. It decides *whether and how* host capabilities are modeled and records the decision the `host:` block never had. Concrete syntax and the eventual removal of the old block are **dependent follow-ups** (named at the end), deliberately not bundled here.

### Problem

The `host:` block (`docker`, `network`, `filesystem`, `privileged`) shipped in v1 **with no design decision**. Confirmed: no `D-*` ratifies it — every `host` mention in `spec/DESIGN.md` lives in the `$`-expression layer (`$host` resource property in D-7; `$app.host` in D-33/D-35; the three-homes litmus in D-36), and the rejection of `host:` as a *top-level field* is in D-33's Rejected list — none of them govern the capability block. It is an unrecorded, half-specified field, and it shows:

1. **It names an implementation at the intent layer (P-1 leak).** `host.docker`, schema text "needs **Docker** daemon access." Move the host to Podman/containerd and the need is identical but the value would still have to say `docker` — fails the platform-agnostic litmus.
2. **Zero catalog adoption; the apps that need it can't use it.** No app declares a `host:` block. `catalog/drafts/portainer`: *"requires the Docker socket … cannot be expressed in a standard Launchfile"* (even though `host.docker` exists). `catalog/drafts/wg-easy`: *"requires NET_ADMIN … or run privileged"* — comment only.
3. **The spec's own example declares the need twice.** `spec/examples/host-orchestrator.yaml` has both `host.docker: required` **and** `requires: - type: docker` ("Orchestrates app containers on the host"). The author reached for both because neither alone can both **gate** ("refuse if you can't grant this") and **wire** (tell the app the granted socket path, which varies by provider). The current `host:` block can gate but not wire; `requires:` already wires (`requires: postgres` → `$url` via `set_env`).

### Why this is a ratification, not just a change

Because `host:` was never ratified, this RFC is the **first decision** about it. That has two consequences worth stating plainly:

- We record the model the field never had — explicitly, not by silently re-deciding an unrecorded field.
- Reshaping or retiring it carries **less P-13 weight** than deprecating a ratified field: it was never a committed part of the contract, so freezing/changing it is lower-risk by construction.

### Proposed model (the boundary)

[...]

PR #170, opened 2026-08-22
spec: host-capability entries — container_runtime syntax, coordinates, grant/refuse (D-44)

### LINKED ISSUE #118: RFC (Follow-up A): host-capability syntax — container_runtime, the host: marker, and wiring
## RFC (Follow-up A): host-capability syntax — `container_runtime`, the `host:` marker, and wiring

Implements the **ratified** host-capability boundary (#113): host capabilities are a **grant/refuse** fulfillment mode of `requires:`/`supports:`. This RFC settles the **concrete syntax** for the first capability and the structural rules that apply to all of them. Scoped tightly to `container_runtime` (the cleanest interface-value case); other capabilities (device, mount, multicast) reuse the same shape and are filed as they're needed.

### Proposed syntax

```yaml
requires:
  - postgres                              # backing service → provision + wire
  - host: { container_runtime: docker }   # capability      → grant or refuse
    set_env: { DOCKER_HOST: $socket }
supports:
  - host: { container_runtime: any }      # optional / enhancing
```

### Ratified rules this syntax must honor (from #113)

- **`host:` marker is required** on every privileged entry — so a privileged request is machine-extractable from the file itself, with zero tooling. An entry's kind: bare string / `type:` = backing service; `host:` = capability.
- **Value-as-interface.** `container_runtime: <interface>` names a control API, not a product: `docker` = the Docker Engine API (a Podman-compatible socket satisfies it), `any` = runtime-agnostic. Open-valued (unknown values tolerated, L-4). **It must not be a product name** — that's the P-1 guarantee the old `host.docker` broke.
- **Required vs optional = `requires` vs `supports`** (extends D-8; Android `uses-feature` / browser-extension `optional_permissions` precedent).
- **Wiring** via `set_env` exposes provider-supplied coordinates — for `container_runtime`: `$socket`, `$host` (`DOCKER_HOST`), `$api` — a D-36 home-#3 value, parallel to `$postgres.url` (D-4/D-7).
- **`validate` audit summary.** `launchfile validate` emits "host capabilities requested: […]" so the privilege surface is visible even before any lint exists. **This RFC's merge is gated on that summary existing.**

### Provider contract

`PROVIDERS.md` gains a **grant/refuse** fulfillment mode for `host:` entries (today it defines only provision for `requires`). A container-based provider mounts/forwards the runtime socket or refuses; a non-container provider refuses or warns. (The reference Docker provider already hard-skips `host.docker: required` — that becomes the documented grant/refuse behavior.)

### Reshape of the existing `host:` keys

`network` / `filesystem` / `privileged` move into the same `{host: {…}}` entry form (they are grant/refuse capabilities too). The legacy `host:` block is normalized into entries — but its **removal** is gated on the legible-evolution principle + the `host.docker` deprecation (filed separately); this RFC adds the new form and keeps the old one working (P-13).

### Motivation (catalog-tested)

`container_runtime`: Diun, Portainer, Dockge, Beszel (Docker socket). `privileged`: wg-easy (NET_ADMIN). The next capabilities — device (Home Assistant USB, Jellyfin `/dev/dri`), host mount (Calibre-Web), multicast (Home Assistant) — are gaps G-9/G-11/G-12 and reuse this shape when filed.

### Scope

`container_runtime` syntax + the universal rules above + the provider-contract update + the `validate` summary. **Out of scope:** the device/mount/multicast values (same shape, filed on demand); resource floors (#113 door-open); the deprecation/removal of the old keys (gated on the legible-evolution principle).

### Open questions

1. Entry nesting: `{host: {container_runtime: docker}}` vs a flatter form.
2. Capability property names: `$socket` / `$host` / `$api`.
3. How strictly `validate` must enforce the required `host:` marker (warn vs error).

ziadsawalha:
Implements the concrete syntax for the host-capability boundary ratified in #113, and records it as **D-44**.

Closes #118

## What lands

A `requires:`/`supports:` entry may now be a **capability** rather than a backing service:

```yaml
requires:
  - postgres                              # backing service → provision + wire
  - host: { container_runtime: docker }   # capability      → grant or refuse
    set_env: { DOCKER_HOST: $socket }
supports:
  - host: { container_runtime: any }      # optional / enhancing
```

Four rules bind every capability entry:

1. **The `host:` marker is required** on every privileged entry — the privilege surface is machine-extractable from the file itself, with zero tooling.
2. **The value names an interface, never a product.** `container_runtime: docker` is the Docker Engine API (a Podman-compatible socket satisfies it); `any` is runtime-agnostic. Open-valued, L-4 style.
3. **Required vs optional = `requires` vs `supports`**, extending D-8.
4. **Wiring reuses `set_env`** (D-4, refined by D-19): a granted capability exposes provider-supplied coordinates `$socket` / `$host` / `$api` — home-#3 values per D-36, parallel to `$postgres.url` (D-7).

`PROVIDERS.md` gains the matching **grant/refuse** fulfillment mode: grant = mount/forward the coordinate and populate the properties; refuse = a clear surfaced message, never a silent drop.

The legacy top-level `host:` block stays valid — P-13. Its deprecation is #120, gated on P-14/D-42.

## How the #118 verdict was addressed

The steward ACCEPTed #118 with five PR-level commitments. Each is honored here:

| # | Commitment | Where |
|---|---|---|
| 1 | `validate` host-capability summary must exist, as a CI-checkable assertion — it was the stated merge gate | `sdk/src/commands.ts` emits `host capabilities requested: […]`; `packages/launchfile/src/__tests__/validate-host-capabilities.test.ts` asserts it (and asserts silence when none are requested) |
| 2 | `PROVIDERS.md` grant/refuse mode must state the conformance obligation — mount-and-populate **or** refuse-with-clear-message, surfaced per §10 | `spec/PROVIDERS.md`; the Docker provider's hard-skip is upgraded from a bare `warnings.push(… skipped)` to a surfaced refusal (`providers/docker/src/compose-generator.ts`, `provider.ts`) |
| 3 | Docs-site gate — `www-dev` does not auto-discover examples; register in **both** `examples/index.astro` and `examples/[slug].astro`, plus a `learn/` mention (the #94 blind spot) | `spec/examples/host-container-runtime.yaml` + both Astro pages + `learn/specialized-patterns.astro` |
| 4 | Carry **D-4/D-19** (not D-10) into the DESIGN.md entry; cite D-36 home-#3 for the coordinates | `spec/DESIGN.md` D-44 |
| 5 | One new `D-*` recording the concrete syntax, referencing #113 as its ratifying parent | D-44, links #113 and #118 in both directions |

## Open questions from #118, resolved

The verdict called all three "spellings, not boundaries" and deferred them to this PR:

1. **Entry nesting** — kept the `{host: {…}}` envelope. A flatter `- container_runtime: docker` loses the machine-extractable privilege marker #113 made the security floor. Recorded under **Rejected**.
2. **Coordinate property names** — pinned `$socket` / `$host` / `$api` in the SPEC table, open-valued for extension.
3. **Marker enforcement strictness** — starts as a lint **warning**; the summary emission is the hard gate, and enforcement can harden additively (P-13). Recorded under **Rejected**.

## Scope

`container_runtime` only. Device / host mount / multicast (G-9/G-11/G-12) reuse this identical shape and are filed on demand. Removal of the legacy keys is #120, not this PR.

## Tests

- sdk 261/261
- providers/docker 89/89
- packages/launchfile 42/42

## Merge order

This is **1 of 9** in the D-44 → D-52 governance merge train. D-44 is the next free decision number on `main`; merging out of order forces every downstream branch to renumber.


ziadsawalha:
Both blockers addressed in `9ba0075`, `ffdb697`, `dc69a4a`. Taking the "fix all the call sites" option rather than narrowing §11, so the equivalence sentence is literally true on merge.

## B2 — mixed `type:` + `host:` entry

Fixed, but **not** with `.strict()`. That was the first suggestion and it breaks the repo: `spec/examples/host-orchestrator.yaml` carries `description:` on a `requires` entry, so closing `RequirementObjectSchema` would invalidate a shipped example (and would collide with the unknown-field-preservation direction).

Instead each union branch now excludes the other's discriminator (`host: z.never().optional()` / `type: z.never().optional()`), so a mixed entry matches neither. A check ahead of the union supplies the message, because a bare union failure only says `Invalid input` for the shape most likely to be written by mistake:

```
an entry is either a backing service (`type:`) or a host capability (`host:`),
never both (D-44) — split it into two entries
```

Four regression tests in `host-capabilities.test.ts`, including one asserting that an unrecognized key on a backing-service entry is *still* tolerated, so nobody re-tightens this later without noticing what it costs.

## B1 — the unguarded call sites

All of them, plus refusal paths:

| Site | Was | Now |
|---|---|---|
| `catalog/test/src/launch-to-compose.ts` | entry form deployed the component, socket ungranted | refuses; both spellings identical |
| `providers/aws/src/translate.ts` | `workaround: no managed AWS service mapping for resource type 'host'`, HCL still emitted | `blocker` for requires, `nice-to-have` for supports — matching the legacy severity |
| `providers/macos-dev/src/provider.ts` | `No provisioner for resource type: host (skipping)`, then launched anyway | refuses with a surfaced message naming the capability |
| `providers/macos-dev/src/env-writer.ts` | correct only by accident (`!props` guard) | explicit |
| `sdk/src/select.ts` | `up host` answered with the synthetic internal type | capabilities excluded from selection |

Verified behaviorally, not just by tests. The harness case that was the most serious finding:

```
entry form  | deploys? false | ["default: requires host capabilities this harness cannot grant (container_runtime=docker) — skipped"]
legacy block | deploys? false | ["default: requires Docker socket — skipped"]
```

Before the fix the first line read `deploys? true`.

## One thing the review did not catch, found while verifying B1

`emitComponent` **returns early** for an image-without-runtime component, above every host check. So on AWS neither spelling reported a capability for a bare-image app — and image-only is exactly the shape of the apps that ask for a runtime socket (Dockge, Portainer). The privilege surface was invisible for the motivating cases, in both spellings equally.

This is pre-existing and not a D-44 regression, but it makes D-44's stated floor false on AWS, so the audit moved above that early return. It does change legacy behavior: a bare-image component with `host.docker: required` now reports `blocker:host.docker` where it previously reported only `workaround:image`. Flagging it explicitly as a deliberate behavior change rather than burying it — if you'd rather it were a separate PR, say so and I'll split it out.

## Non-blocking notes

Not addressed here, and I'd rather track them than quietly fold them in — happy to take any into this PR if you'd prefer:

- **N1** (`$host` collides with D-7's hostname meaning) — a real naming call. It wants an Author decision, not a unilateral rename, since it changes ratified spec surface.
- **N3** (`KNOWN_HOST_CAPABILITIES` omits `docker`) — agreed, and your point that it would flag the repo's own `host-orchestrator.yaml` is the argument for adding it. That example should migrate to the entry form.
- **N6** (nothing validates the published JSON Schema; `host-orchestrator.yaml` is already invalid against it) — a genuine `P-5` conformance gap, worth its own issue.
- **N2, N4, N5, N7, N8** — N4 is fixed as part of B1. The rest stand as filed.

Tests: sdk 265/265, providers/docker 89/89, providers/macos-dev 95/95, providers/aws 38/38, packages/launchfile 42/42, catalog/test 10/10.

ziadsawalha:
Fair catch on B1-residual, and an uncomfortable one — I fixed the message and not the behavior, which is exactly the failure mode that's hardest to spot once the message reads correctly. Confirmed at `provider.ts:157-178`: the block only called `console.error`, and nothing downstream consulted it.

Addressed in `7eb62e9` and `4aa1fd0`.

## B1-residual — macos-dev now actually declines

A refused component is removed from the component map **before** anything downstream runs, reusing the same mechanism the component selector already uses, so the runtime-install, env-write, `pm2.register` and `startAll` loops never see it. No `Set` threaded through eight loops.

Both spellings fold into one check, which also closes the equivalence gap you flagged — the legacy block previously produced *no output at all* here, since nothing in this package read `comp.host`.

Verified by running a real launch, mixed app, one refused component and one plain:

```
entry form:
  Refused: needsdocker requires host capabilities this provider cannot grant (container_runtime=docker) — component not started
  ↓ Starting services...  [plain] Starting: sh -c 'echo PLAIN_ONE_STARTED; sleep 5'
  refused comp started? no (correct)
  plain comp started?   yes (correct)

legacy block:
  Refused: needsdocker requires host capabilities this provider cannot grant (container_runtime=docker (host.docker)) — component not started
  refused comp started? no (correct)
  plain comp started?   yes (correct)
```

And the single-component case now exits instead of reporting success over an empty set:

```
  Refused: default requires host capabilities this provider cannot grant (container_runtime=docker) — component not started
Every selected component requires a host capability this provider cannot grant.
```

No log file is written in either case.

I took the **decline** option rather than **grant**. Granting a Docker socket is defensible on a dev box, but it would mean this provider silently acquiring a privilege on the user's behalf, and it needs `$socket`/`$host` resolution that no provider implements yet — a bigger change than this PR should carry. N2 stays open.

**On testability:** you were right that a message assertion would have passed. The check is extracted as `refusedHostCapabilities()` and the seven new tests assert *outcomes* — which components are refused, that both spellings produce identical sets, that an optional capability is not refused, and that a refused component doesn't take its siblings down. macos-dev went from 95 to 102 tests.

## B2-new — conformance report

Regenerated. `bun run conformance --check` now passes. 83 → 84 files, 85 → 87 gaps, exactly as you measured.

## N10 — published schema

Taken, as you suggested. `www-dev/public/schema/v1` was three decisions behind: no `source` (D-38), no `capture` (D-34), and still carrying the removed top-level `outputs`.

Verified with ajv after the sync: **11 of 12 spec examples and 71 of 72 catalog apps validate.** The two exceptions are `host-orchestrator.yaml` and `remote-claude-concentrator/Launchfile`, and I checked they are already invalid against the schema published *today*, for the same `additionalProperties: false` reason before and after. So this is a strict improvement — `remote-claude-concentrator` additionally fails today on `commands/bootstrap`, which the sync fixes.

I deliberately did **not** open `requirement` or `storageVolume` to fix those last two. That is the ratification deferred on #171, and it is not mine to make unilaterally inside a D-44 PR.

## Not taken, deliberately

- **The image-only + legacy-block test** you asked for — added? No. Flagging honestly: I added the entry-form image-only test but not the legacy-block one, and you're right that the legacy case is the one that actually changed behavior. Say the word and I'll add it; I'd rather you know it's missing than assume it's there.
- **N3** (`docker` missing from `KNOWN_HOST_CAPABILITIES`) — still open. Adding it flags the repo's own `host-orchestrator.yaml`, which I think is correct, but it turns a shipped example into a warning and that felt like it wanted a nod first.
- **N1** (`$host` vs D-7) — Author decision, unchanged.
- **N11, N12, N13** — noted, not addressed here.

Tests: sdk 265/265, providers/docker 89/89, providers/macos-dev 102/102, providers/aws 38/38, packages/launchfile 42/42, catalog/test 10/10.

ziadsawalha:
B3 addressed in `62a514d`.

You were right that the seven tests exercised the classifier, not the wiring — I checked your counterfactual myself before fixing it, and it reproduced exactly: gutting the removal left all 102 passing.

The removal step is now `applyHostCapabilityRefusals()`, called by `launchUp`, so a test drives the same code path rather than a copy of it. Seven tests assert outcomes: the refused component is absent from what would reach the process manager, siblings still start, the component map itself shrinks, both spellings behave alike, and `none-left` is reported when nothing survives.

**Verified the fence bites.** Reinstating the defect — refusal printed, component left in the map — now fails 5 tests where it previously failed 0:

```
with the bug reinstated:   104 pass  5 fail
restored:                  109 pass  0 fail
```

Also filed **#179** for N12, since you flagged it as the highest-value follow-up and asked for it before merge: one CI job covering `conformance --check`, the `www-dev/public/schema/v1` sync, and ajv validation of `spec/examples/**`. All three classes in this PR were caught by hand — twice — and the schema copy had already shipped two decisions with the gap open.

On **N16** (the scope asymmetry — legacy block is app-scoped, entry form component-scoped): agreed that it reads like a bug and isn't. I've left §11 alone rather than editing normative text late in review, but it's a fair sentence to add.

**N5** (the synthetic `type: "host"` that a discriminated union would have made a compile error) I'd also rather see land before the surface hardens — but as its own change, not bolted onto this one.

Tests: sdk 265/265, providers/docker 89/89, providers/macos-dev **109/109**, providers/aws 40/40, packages/launchfile 42/42, catalog/test 10/10.

ziadsawalha:
Correction to the test counts in my previous comment: **sdk is 270/270**, not 265 — I quoted a figure from before the lint advisory tests landed. The rest were right.

Verified just now on `62a514d`: sdk 270/270, providers/docker 89/89, providers/macos-dev 109/109, providers/aws 40/40, packages/launchfile 42/42, catalog/test 10/10.

Issue #171, opened 2026-08-22
RFC: unknown fields are preserved — pin down what P-13's "ignore" obliges


ziadsawalha:
### Problem

`P-13` promises *"a v1 parser ignores unknown fields gracefully"*, and `SPEC.md` §Extensibility (lines 1105–1106) promises:

> - New fields can be added at any level without breaking existing files
> - Unknown fields are ignored by parsers that do not support them

**"Ignore" is currently implemented three different ways, and they contradict each other and the spec.** All three were verified against `main`, not assumed:

**1. The JSON Schema rejects unknown fields below the root.** The root object is open (no `additionalProperties`), but **12 of 23 `$defs` are `additionalProperties: false`** — `secret, provides, requirement, envVar, build, health, commandDetail, output, storageVolume, dependsOnEntry, component, host`.

Validating a minimal document against `spec/schema/launchfile.schema.json` on `main` (ajv, draft 2020-12):

```
doc: { version: launch/v1, name: gitea, components: { web: { image: nginx, x-future-field: nested } } }

valid: false
  /components/web must NOT have additional properties {"additionalProperty":"x-future-field"}
```

This directly contradicts `SPEC.md:1027`, which documents `x-` as *"a convention for extension fields that parsers ignore"* — the convention is invalid the moment it appears one level below the root.

**2. The reference SDK silently strips them.** Round-tripping a real catalog shape (`catalog/apps/gitea/Launchfile`, plus two extension fields and a `size:`) through `readLaunch` → `writeLaunch` on `main`:

```yaml
# in
version: launch/v1
name: gitea
image: gitea/gitea:latest
x-vendor-note: top-level-extension
requires:
  - type: postgres
    x-tuning: nested-in-requirement
storage:
  data: { path: /data, persistent: true, size: 10Gi }
```
```yaml
# out
version: launch/v1
name: gitea
image: gitea/gitea:latest
requires:
  - postgres
storage:
  data: { path: /data, persistent: true }
```
```
x-vendor-note    *** LOST ***
x-tuning         *** LOST ***
size             *** LOST ***
```

Note `size: 10Gi` is **not** an unknown field — it is `D-30`'s storage size hint, present in the JSON Schema and dropped by the SDK. Same class of drift, already-ratified field.

**3. So "ignore" means allow (root), reject (`$defs`), and strip (SDK)** — three different contracts for the same input. `P-5` (every parser and provider translates the same file the same way) cannot tolerate that.

[...]

PR #176, opened 2026-08-22
spec: resource property registry — vocabulary is standard but open (D-46)

### LINKED ISSUE #172: RFC: resource property registry — the vocabulary is standard but open (delivers L-4, closes G-6)
## RFC: resource property registry — the vocabulary is standard but **open**

Delivers the **"Future"** half of limitation **L-4** and closes catalog gap **G-6**. Both are already filed in this repo; this proposal is the thing they each point at.

### Problem

`SPEC.md` § Resource Property Vocabulary documents a standard set of resource properties — `url`, `host`, `port`, `user`, `password`, `name` — as **convention**. Nothing checks them.

**L-4** (`spec/DESIGN.md`) states the problem and names the fix:

> **Problem**: The standard properties (`url`, `host`, `port`, `user`, `password`, `name`) are convention, not enforced by the schema. A typo like `$hoost` passes validation and silently resolves to an empty string.
> **Future**: Build a machine-readable property registry per resource type with post-parse validation.

**G-6** (`catalog/GAPS.md:38–42`) files the same thing from the catalog side:

> **Apps**: All apps using `set_env`
> **Issue**: `$host`, `$port`, `$url` etc. are convention, not validated. A typo like `$hoost` isn't caught.
> **Suggestion**: Define a machine-readable property registry per resource type (already in plan as L-4).

Verified on `main` with G-6's own example — a `postgres` requirement wiring `DB_HOST: $hoost`:

```
warnings: (none)
```

The typo is accepted silently and resolves to `""` at runtime, so the app starts and then fails somewhere unrelated with an empty host.

### Proposed change

Ratify the vocabulary as a **registry that is authoritative but not closed**, plus an advisory lint.

1. **The vocabulary for a known type is OPEN.** A provider extension is legitimate; an unrecognized property is *not* an error. `SPEC.md` already permits provider-specific properties and that stays true.
2. **Semantics notes are descriptive only.** A note records the documented meaning of a property; it confers no validation behavior.
3. **One vocabulary, consistency-checked.** The `SPEC.md` prose table stays canonical; `spec/schema/resource-properties.json` is its machine-readable mirror, and a test keeps the two from drifting.
4. **A lint warns on non-standard properties for a *known* type.** Unknown resource types stay fully open (L-4) and silent.

After the change, the same probe:

```
warnings: [
 "\"$hoost\" is not in the standard vocabulary for postgres (known: url, host, port, user, password, name)"
]
```

**The severity is a warning, not an error** — verified: `validate` still returns `valid: true` with warnings attached (`sdk/src/commands.ts:154–163`). An open vocabulary that hard-fails is a closed one, so this is load-bearing, not a detail.

### Motivation

Being straight about the catalog evidence, because it cuts both ways:

- **29 of 72 catalog apps use `set_env`** with resource-property references. Across the catalog the standard vocabulary is used heavily: `$host` ×19, `$name` ×18, `$user` ×17, `$url` ×17, `$password` ×17, `$port` ×14.
- **Running the proposed lint across all 72 apps produces 0 warnings.** The catalog is currently clean. This lint fixes nothing today — it is preventive.

So the honest case is not "the catalog is broken". It is:

1. **G-6 and L-4 are already-accepted statements that this gap is worth closing** — both filed in-repo, both naming a property registry as the fix. This proposal is the delivery, not a new argument.
2. **The exposure is 29 apps and every future one.** The failure mode is a silent empty string, which surfaces as an unrelated runtime error far from the typo.
3. **A registry is a prerequisite for anything better later** — editor completion, `$prop` hover docs, provider conformance checks. None of that can exist while the vocabulary lives only in prose.

If the steward considers a preventive lint with zero current hits below the bar, DEFER is a fair call — though L-4 and G-6 would then both stay open with their named fix declined, which is worth stating explicitly if so.

### Principle self-assessment

| Principle | Assessment |
|---|---|
| **P-1** app-focused | Neutral. No new app-facing field. |
| **P-3** machine-generatable | **Strengthened.** A generator can target a machine-readable vocabulary instead of prose. |
| **P-4** human-writable | **Strengthened.** A typo gets a message instead of a silent empty string. |
| **P-5** provider-translatable | **Strengthened.** One published vocabulary, consistency-checked against the prose table. |
| **P-6** it's just YAML | **Aligned.** No syntax change; lint-only. |
| **P-13** additive extensibility | **Preserved — and this is the crux.** The registry is authoritative but open. Closing it would violate P-13; warning does not. |
| **L-4** | **Delivers its stated Future**, keeping its "vocabulary is open" stance intact. |
| **D-7** resource properties as standard vocabulary | Extends D-7 with a machine-readable mirror; changes no property meaning. |

**Risks:**

1. **Warning fatigue / false positives.** A provider that legitimately exposes an extra property triggers a warning on every lint. Mitigation is the wording (`"not in the standard vocabulary"`, advisory) — but the noise is real if a provider extension becomes common.
2. **Two sources of truth.** The prose table and the JSON mirror can drift. Mitigated by a consistency test, which is itself a thing that can be deleted.
3. **Reversibility: high.** Lint-only; removing it changes no file's meaning.

### Open question for review

Should the lint stay silent for a **known type with no `set_env`** but a `$prop` referenced elsewhere (e.g. in `env` or a command)? The proposed check only walks `set_env` on requirements/supports. That is the common case, but it is not all of them — extending it later is additive.

ziadsawalha:
Ratifies the resource property vocabulary as a registry that is **authoritative but open**, and adds an advisory lint that catches typos without forbidding extension. Records it as **D-46**.

Closes #172

## What lands

`SPEC.md` § Resource Property Vocabulary already listed the standard properties (`url`, `host`, `port`, `user`, `password`, `name`). Nothing checked them, so `$hoost` validated cleanly and resolved to an empty string at runtime — the app then failed somewhere unrelated.

Four rules:

1. **The vocabulary for a known type is OPEN.** A provider extension is legitimate; an unrecognized property is not an error.
2. **Semantics notes are descriptive only** — a note records meaning, never validation behavior.
3. **One vocabulary, consistency-checked.** The `SPEC.md` prose table stays canonical; `spec/schema/resource-properties.json` mirrors it, and a test parses the prose as ground truth so the two cannot drift.
4. **A lint warns for a *known* type.** Unknown types stay fully open (L-4) and silent.

Before / after, using G-6's own example:

```
before:  warnings: (none)
after:   warnings: ["\"$hoost\" is not in the standard vocabulary for postgres
                     (known: url, host, port, user, password, name)"]
```

**The severity is a warning, not an error** — `validate` still returns `valid: true` and exits 0. An open vocabulary that hard-fails is a closed one, so this is load-bearing rather than a detail.

## Delivers work already planned

- **L-4**'s stated *Future*: "Build a machine-readable property registry per resource type."
- **G-6** in `catalog/GAPS.md`, whose Suggestion names the same registry. Marked resolved here.

## How the #172 verdict was addressed

The steward returned **ACCEPT / High confidence, zero blocking items**, with notes. Two are folded into this PR:

| Note | Addressed |
|---|---|
| **1 (highest priority)** — SPEC.md grants only *type* openness; *property* openness for a known type existed only for `$app.*`. A reader of SPEC.md alone could not tell whether a warning meant their file was wrong. | `spec/SPEC.md` now states the openness rule normatively, and says why the diagnostic is advisory: the property may be a typo *or* a deliberate provider extension, and the spec cannot tell them apart. |
| **2** — D-7's *Rejected* clause reads "explicit property declarations per resource type", which a fast reader will mistake for this. | D-46's Why now records the distinction: D-7 declined *author-side* declarations in the app file; this publishes the same convention *spec-side*, with no author-facing surface. |

The remaining notes (3–7) are tracked, not silently dropped — see below.

## One thing the verdict found that the proposal understated

The RFC claimed the lint is purely preventive (0 warnings across 72 apps — true). The steward showed it also has present-tense discriminating power against the shipped reference providers:

| Type | Provider exposes | SPEC table | Lint on correct usage |
|---|---|---|---|
| `elasticsearch` | `…, user, password, name` | `url, host, port` | 3 warnings |
| `minio` | `…, bucket, region` | no `region` | 1 warning |
| `sqlite` (macos-dev) | `url, path, host, port` | `url, path` | 2 warnings |

Under D-33's posture (*"portable Launchfiles should use only the standard set"*) those warnings are **correct** — they say the file is provider-dependent, not portable. Whether the SPEC table should widen instead is a separate spec change with its own argument, deliberately not made here (rule 2 of this decision). Tracked as a follow-up.

## Stacking

Based on `chore/gov-host-syntax` (**D-44**), not on the D-45 branch: D-45 (unknown-field preservation) was **deferred** on #171 and its number stays reserved for the reworked proposal. So `DESIGN.md` reads D-44 → D-46 with D-45 vacant by intent. Merge D-44 first.

## Tests

- sdk 280/280
- providers/docker 89/89, providers/macos-dev 102/102, packages/launchfile 42/42, catalog/test 10/10
- All `DESIGN.md` / `SPEC.md` / `PROVIDERS.md` anchors resolve
- 113 catalog files linted: **0 warnings**, so no false positives on shipped entries


ziadsawalha:
Blocker fixed in `7be21dc`, and note 4 went in with it. CI has now run on this PR — all 12 checks green.

## The fix

`Object.hasOwn` at both call sites, and a null prototype on the exported constant:

```ts
const vocabulary = Object.hasOwn(RESOURCE_PROPERTY_VOCABULARY, req.type)
  ? RESOURCE_PROPERTY_VOCABULARY[req.type]
  : undefined;
```

Three things, deliberately:

1. **`lint.ts:112`** — the crash. This is the blocker.
2. **`lint.ts:155`** (`PRODUCT_SPELLINGS[req.type]`) — note 4. You called it for #170 because that is where the line is anchorable, but #170 merged before this got addressed, so the line now lives on `main` and there is no open PR to anchor it to. It is the same one-line bug, the same shape, three lines away in the same function. Splitting it out to chase a merged PR would have cost more than it saved. Fixed here.
3. **`resource-properties.ts`** — `RESOURCE_PROPERTY_VOCABULARY` gets a null prototype. Your point stands that the call-site guard does nothing for downstream consumers, and the constant is exported from `sdk/src/index.ts`, so any `V[type]` lookup in a consumer inherits the trap. The guard unblocks the PR; the null prototype is what stops it being re-introduced by the next caller.

`Object.hasOwn` is the durable half of that pair. The reason TypeScript could not see the original bug — `Record<string, readonly string[]>` types the inherited hit as a valid entry — applies just as much to the *next* prototype lookup someone writes, and a type annotation will keep lying about it. The runtime guard does not.

## Verification — all OBSERVED

Both ends reproduced directly, in throwaway worktrees at each commit, not inferred from the test.

**Before (`a7da567`)** — all seven `Object.prototype` keys crash:

```
type=constructor           -> exit=1 ::   → vocabulary.includes is not a function
type=__proto__             -> exit=1 ::   → vocabulary.includes is not a function
type=toString              -> exit=1 ::   → vocabulary.includes is not a function
type=valueOf               -> exit=1 ::   → vocabulary.includes is not a function
type=hasOwnProperty        -> exit=1 ::   → vocabulary.includes is not a function
type=isPrototypeOf         -> exit=1 ::   → vocabulary.includes is not a function
type=propertyIsEnumerable  -> exit=1 ::   → vocabulary.includes is not a function
```

**After (`7be21dc`)** — same seven files, same `$host` property:

```
type=constructor           -> exit=0 :: ✓ acme is valid
type=__proto__             -> exit=0 :: ✓ acme is valid
type=toString              -> exit=0 :: ✓ acme is valid
type=valueOf               -> exit=0 :: ✓ acme is valid
type=hasOwnProperty        -> exit=0 :: ✓ acme is valid
type=isPrototypeOf         -> exit=0 :: ✓ acme is valid
type=propertyIsEnumerable  -> exit=0 :: ✓ acme is valid
```

Your report named five keys; `isPrototypeOf` and `propertyIsEnumerable` fail identically, so the new suite pins all seven rather than the five that were reported.

| Check | Result |
|---|---|
| `bun test` (sdk) | **287 pass, 0 fail** — was 280; +7, one per prototype key |
| `bun run typecheck` | clean |
| CI on this PR | **12/12 pass**, including `AWS provider — terraform validate` |

## Your note 5 is now retired

> No CI has run on this PR. […] the merge prerequisite "CI passes" can't be met until #170 lands and this retargets `main`.

That is exactly what happened. #170 merged, GitHub retargeted this PR to `main`, and the push of `7be21dc` triggered the full matrix for the first time. `mergeStateStatus` is now `CLEAN`.

Worth flagging for the rest of the stack: **#193, #194, #196 and #197 still report 0 checks each** — verified, not assumed:

```
$ for pr in 193 194 196 197; do gh pr view $pr --json statusCheckRollup --jq '.statusCheckRollup|length'; done
0
0
0
0
```

They inherit the same cause, and the retarget-on-merge trick only helps whichever one is next in line. **#195** drops the `branches: [main]` filter and fixes it for all of them at once. It is green and `CLEAN`, but it needs an independent verdict — the steward drafted it and correctly declined to review its own work.

## Merge-order check

Since this PR now changes `sdk/src/lint.ts` and four downstream branches also touch that file, I checked whether landing this forces a rebase cascade. It does not:

```
$ for b in gov-failure-semantics gov-provenance schedule-honesty required-env; do
    git merge-tree $(git merge-base 7be21dc origin/$b) 7be21dc origin/$b | grep -c '^<<<<<<<'
  done
0
0
0
0
```

The downstream edit adds `lintDurations` to a different hunk. The train merges cleanly in order.

## Not addressed here, on purpose

The missing **D-45** record. `git grep D-45` returns nothing — the reservation for #171 exists only in PR prose, and a reader of a normative numbered list hits an unexplained hole at D-44 → D-46. You raised it on #177, but it is introduced by *this* PR and becomes permanent the moment #177 lands on top. It belongs here, before this merges. Tell me which you want — a one-line reservation stub in `DESIGN.md`, or a tracked issue — and I will add it to this branch.

ziadsawalha:
Applied your D-45 suggestion in `62bc959` — taking the stub over the issue, for the reason you gave: an issue records that the document is wrong, it doesn't make it right, and `www-org` renders all of `DESIGN.md` either way.

Verified before applying: `D-1`…`D-44` are contiguous, so this is the file's only gap, and it is genuinely introduced here — `origin/main` ends at D-44 and this PR adds D-46.

One correction to my own earlier reply, since it is on the permanent record: I wrote that the two unguarded lookups were "three lines away in the same function". They are in **different** functions — `checkResourceProperties` and `lintLaunch` — about 46 lines apart. The scope argument for fixing both here is unaffected; the premise I gave for it was wrong.

Worth flagging that the same hole exists one number up: `DESIGN.md` at the tip of this train reads D-49 → **D-51** → D-52. Verified across every branch in the stack —

```
$ for b in main gov-resource-registry gov-secret-generator gov-failure-semantics \
           gov-provenance schedule-honesty required-env; do
    git show "origin/$b:spec/DESIGN.md" | grep -cE '^#+ D-(45|50)\b'
  done
0 0 0 0 0 0 0
```

— and neither number appears anywhere in `DESIGN.md` or `PROVIDERS.md` prose on any of them. D-45 is closed by this commit. **D-50 arrives from #196** and becomes permanent when #197 lands on top of it, so it wants the same treatment there. Flagging it here rather than filing, since you own the call on both.

PR #177, opened 2026-08-22
spec: define generator: secret output — 32 bytes, hex-encoded, lowercase (D-47)

### LINKED ISSUE #174: RFC: define what `generator: secret` produces — 32 bytes, hex-encoded
### Problem

`SPEC.md` says `generator: secret` produces a *"cryptographically random hex string"*. Charset named, **length unspecified** — so any length conforms, and the three in-tree providers each picked their own:

| Provider | Output on `main` | Source |
|---|---|---|
| `@launchfile/docker` | 32 bytes, hex → 64 chars | `providers/docker/src/compose-generator.ts:50-54` |
| `@launchfile/macos-dev` | 32 bytes, **base64url** → 43 chars | `providers/macos-dev/src/secret-generator.ts:19` |
| `@launchfile/aws` | 32 **alphanumeric characters** (`random_password`, `special = false`) | `providers/aws/src/translate.ts` |

Two of the three do not even emit hex, so they contradict the shipped wording. One Launchfile, three different secrets, three different strengths, depending on who runs it. That is the `P-5` failure — the same file must translate the same way everywhere.

### This is not theoretical — it breaks three shipped catalog apps today

`firefly-iii`, `monica` and `snipe-it` are Laravel apps. Each writes:

```yaml
secrets:
  app-key:
    generator: secret
env:
  APP_KEY: "base64:${secrets.app-key|base64}"
```

Laravel requires `APP_KEY` to decode to **exactly 32 bytes** for AES-256-CBC. The `|base64` transform (`sdk/src/resolver.ts:386-395`) decodes a hex input to raw bytes *before* encoding, so hex in gives 32 bytes out. Feed it something that is not hex and it base64s the characters instead.

Ran against the resolver on `main`:

```
docker    (32B hex)            -> APP_KEY decodes to 32 bytes  OK
macos-dev (32B base64url)      -> APP_KEY decodes to 43 bytes  *** Laravel REJECTS (needs 32) ***
aws       (32 alnum chars)     -> APP_KEY decodes to 32 bytes  OK
```

**So all three Laravel apps in the catalog fail to boot on `@launchfile/macos-dev` today** — a 43-byte key, rejected at startup. The aws case lands on 32 bytes by coincidence (32 ASCII characters happen to be 32 bytes), not by contract.

That is the concrete cost of leaving the length undefined: a silent, provider-dependent breakage in shipped catalog entries.

### Proposed change

`generator: secret` produces **32 bytes of cryptographically random data, hex-encoded (64 characters)**.

One sentence of normative prose. No YAML change, no new field, no schema change.

It matches:
- the reference docker provider's existing behavior (a no-op ratification there),
- the `openssl rand -hex 32` convention this ecosystem already uses,
- every documented catalog requirement: outline's hex-encoded 32-byte key exactly, rallly's ≥32-character minimum at 2× margin, and the three Laravel apps' 32-byte key via the existing `|base64` transform (D-32).

The two providers whose output changes (`macos-dev`, `aws`) were **already non-conforming** to the shipped "hex string" wording, so this is not taking a working behavior away from them.

### Motivation

- **29 `generator: secret` declarations across 26 catalog apps.**
- **3 apps actively broken on one provider today** (`firefly-iii`, `monica`, `snipe-it` — verified above).
- Resolves the length half of **G-16**, whose current text (`catalog/GAPS.md:105-108`) says *"a 32-char hex string works for most cases"* — itself inaccurate: the reference provider emits 32 **bytes**, which is 64 hex characters. This PR corrects that claim too.

### Principle self-assessment

| Principle | Assessment |
|---|---|
| **P-1** app-focused | **Pass.** Key size is app knowledge and does not change with deployment target. |
| **P-5** provider-translatable | **The core fix.** Closes three incompatible outputs from one spec sentence. |
| **P-6** it's just YAML | **Pass.** Prose only; zero syntax change. |
| **P-9** unambiguous | **Pass.** No exceptions — `secret` means 32 bytes hex, everywhere. |
| **P-13** additive extensibility | **Pass on its own terms, and the terms matter.** `P-13` is a *format* principle — no field changes meaning, every existing Launchfile stays valid. That holds. **Correction (steward review):** an earlier draft of this line claimed "nothing stored anywhere becomes invalid." That is false and is withdrawn. Secrets *are* persisted — in `.launchfile/state.json` on macos-dev and in Terraform state on aws. See the migration note below. |
| **D-32** pipe transforms | Relies on it rather than changing it — and hex is what makes `\|base64` yield a real 32-byte key. |

### Migration — what happens to already-deployed instances

Added after the steward review, which correctly caught that the original `P-13` line overstated the case.

- **`macos-dev` reuses persisted secrets.** `generateSecrets` skips any name already present in `.launchfile/state.json` (`providers/macos-dev/src/env-writer.ts:176-189`, `if (!secrets[name])`). So an existing deployment keeps its old base64url value — good for compatibility, but it also means **the fix is not retroactive**: the three Laravel apps already deployed on macos-dev stay broken until their state is destroyed.
- **`aws` destroys and recreates.** The change swaps `random_password` for `random_bytes`. That is a different Terraform resource type, so `moved` blocks cannot bridge it and **any existing stack silently rotates its secret on the next `apply`** — unrecoverable for anything already encrypted at rest under the old key. This needs a changeset, a version bump, and an explicit upgrade note, not silence.

**Risks:**

1. **Secrets change value on `macos-dev` and `aws` after upgrade.** For an app that already stored a generated secret in a database (a session key, an encryption key), regenerating it invalidates existing data. Worth calling out in the changeset even though secrets are per-instance.
2. **A 64-character hex string is a poor value for a human to type.** `pihole`'s `WEBPASSWORD` and `photoprism`'s `PHOTOPRISM_ADMIN_PASSWORD` are admin passwords a person may enter by hand. This proposal does **not** solve that; see below.
3. **Reversibility: moderate.** Once providers conform, changing the length again is another conformance break — though still not a file-format break.

### Explicitly out of scope

Per-secret `length` / `encoding` / charset fields. `D-32` already rejected field-level encoding, and with a defined default no current catalog app needs them. The real demand signal is human-typed admin passwords (risk 2) — a separate proposal that must reconcile with D-32's rejection first, rather than being smuggled in here.

### Open question — answered

**Yes: the spec should mandate lowercase hex.** Should `SPEC.md` state the generated value is **lowercase** hex? The reference provider emits lowercase (`toString(16)`), and `|base64`'s hex detection accepts either case — but two providers emitting different cases would still be a `P-5` divergence for anything doing a string comparison.

ziadsawalha:
Defines what `generator: secret` produces — **32 bytes of cryptographically random data, hex-encoded as 64 lowercase characters** — and brings both non-conforming providers to it. Records it as **D-47**.

Closes #174

## The problem

`SPEC.md` said `generator: secret` produces a *"cryptographically random hex string"*. Charset named, **length unspecified** — so any length conformed, and the three in-tree providers each picked their own. Two of the three did not emit hex at all:

| Provider | Output before | |
|---|---|---|
| `@launchfile/docker` | 32 bytes → 64 hex chars | conforming |
| `@launchfile/macos-dev` | 32 bytes → 43 **base64url** chars | non-conforming |
| `@launchfile/aws` | 32 **alphanumeric** chars | non-conforming |

One file, three secrets, three strengths, depending on who ran it. That is the P-5 failure.

## It was breaking three shipped catalog apps

`firefly-iii`, `monica` and `snipe-it` derive a Laravel `APP_KEY` through the `|base64` transform, which requires exactly 32 bytes for AES-256-CBC. `sdk/src/resolver.ts:386-395` hex-decodes before encoding, so hex in gives 32 bytes out — and anything else does not:

```
docker    (32B hex)        -> APP_KEY decodes to 32 bytes  OK
macos-dev (32B base64url)  -> APP_KEY decodes to 43 bytes  *** Laravel REJECTS ***
aws       (32 alnum)       -> APP_KEY decodes to 32 bytes  OK (by coincidence)
```

The macos-dev failure is **deterministic, not probabilistic**: base64url of 32 bytes is always 43 characters, and 43 is odd, so it can never satisfy the resolver's even-length hex gate.

## Why 32 bytes hex is the right value — observed, not argued

`catalog/test/src/launch-to-compose.ts` already generates exactly this, and **25 of the 26** `catalog/apps/` entries that declare `generator: secret` carry `health_check_passed: true` under it — the 26th, `remote-claude-concentrator`, ships no `test:` block at all, so it is untested rather than failed. A further **23** `catalog/drafts/` entries declare it; 2 of those (`anythingllm`, `n8n`) are harness-tested green and the remaining 21 sit outside that evidence.

It also matches outline's documented `openssl rand -hex 32`, clears rallly's ≥32-character minimum at 2×, and is a verified no-op for the reference docker provider.

## How the #174 verdict was addressed

**ACCEPT / High confidence, zero blocking items.** Three substantive corrections landed:

| | |
|---|---|
| **Mandate lowercase** — the verdict answered the RFC's own open question. Leaving case open would reproduce the same divergence one layer down: two conforming providers, different values, and any string comparison or hash lookup disagreeing. | Done. Costs nothing — all three implementations already emit lowercase. |
| **The `P-13` reasoning was wrong.** The issue claimed "nothing stored anywhere becomes invalid." False — secrets *are* persisted. | Withdrawn and corrected in the issue body, with a migration section. See below. |
| **D-32 was over-cited as a fence.** Its ratio (encoding does not compose with interpolation prefixes) covers a future `encoding:` field and says nothing about `length:`. | Both stay deferred — on demand, not on precedent — so a future proposer is not sent to argue against a decision that never covered their field. |

## Migration hazards — read the changeset

Both providers change observable output, and the two hazards point in opposite directions:

- **`aws` destroys and recreates.** `random_password` → `random_bytes` is a Terraform *resource-type* change, which a `moved` block cannot bridge, so an existing stack rotates its secret on the next `apply`. Anything encrypted at rest under the old value becomes undecryptable. **Back up or re-key before upgrading.** Tracked separately as a follow-up so it gets a real migration path rather than only a warning.
- **`macos-dev` is not retroactive.** Generated secrets persist in `.launchfile/state.json` and are reused (`env-writer.ts:176-189`), so existing deployments keep their old value — deliberate, since rotating a live secret on upgrade would be worse. The three Laravel apps already deployed there stay broken until re-provisioned.

## Tests

- sdk 280/280, providers/macos-dev 102/102, providers/aws 40/40
- `registerSecret()` verified surviving the rebase — dropping it would silently revert the redaction fix from #148

## Stacking

Based on `chore/gov-resource-registry` (**D-46**). D-45 is deferred (#171) and its number stays reserved, so `DESIGN.md` reads D-44 → D-46 → D-47.


ziadsawalha:
Filed the AWS destroy-and-recreate hazard as #178 so it gets a real migration path rather than only a changeset warning.

This PR still ships the warning, and D-47 itself is worth landing on its own merits — the old aws output was non-conforming and the divergence it created is a reproduced bug, not a theoretical one. But `random_password` → `random_bytes` is a Terraform resource-type change that `moved` cannot bridge, so an existing stack rotates its secret on the next `apply`, and a key that encrypted data at rest takes the data with it. #178 covers making that transition either lossless or loud.

ziadsawalha:
Both blockers addressed — one fixed here, one now actually reachable.

## Blocker 1 — D-47's catalog evidence — fixed in `6a9560a`

I re-derived the counts from scratch rather than taking them from the review. They reproduce exactly:

```
$ git grep -l 'generator: secret' -- catalog/apps/   | wc -l
26
$ git grep -l 'generator: secret' -- catalog/drafts/ | wc -l
23
$ for f in $(git grep -l 'generator: secret' -- catalog/apps/); do
    d=$(dirname $f); grep -q 'health_check_passed: true' $d/metadata.yaml || echo "$d"
  done
catalog/apps/remote-claude-concentrator
$ grep -c '^test:' catalog/apps/remote-claude-concentrator/metadata.yaml
0
```

25 of 26, and `remote-claude-concentrator` has no `test:` block at all — untested, not failed. Confirmed.

The **Why** paragraph now reads:

> …25 of the 26 `catalog/apps/` entries that declare it carry `health_check_passed: true` under that output (the 26th, `remote-claude-concentrator`, ships no test results at all) — so the harness-tested catalog has already been validated against it; a further 23 `catalog/drafts/` entries declare it and sit outside that evidence.

The conclusion is unchanged, because it survives the correction: 25 harness-tested apps under exactly this output is still the strongest evidence a spec decision in this repo carries. What changed is that a future reader re-running the count now finds what the sentence says.

## Blocker 2 — CI — the sequencing you named is now reachable

Confirmed still true at this head:

```
$ git show HEAD:.github/workflows/ci.yml | sed -n '3,5p'
on:
  pull_request:
    branches: [main]
$ gh pr checks 177
no checks reported on the 'chore/gov-secret-generator' branch
```

Two things have moved since the review:

1. **#170 has merged.** #176 auto-retargeted to `main`, so the chain the unblock path depends on has started.
2. **#195 is open, green, and is the general fix** — it drops the `branches: [main]` filter so CI runs on every PR, stacked or not. 12/12 checks pass, `mergeState: CLEAN`. The steward posted verification evidence on it but explicitly declined to render a verdict, because the steward drafted it. That is the right call, and an independent verdict is being obtained.

So there are two routes to satisfying this blocker, and #195 is the better one: the retarget-and-push route fixes CI for *this* PR, while #195 fixes it for the whole stacked queue — #193, #194, #196 and #197 all have the identical empty `statusCheckRollup` for the identical reason. I'd rather land the general fix than repeat the manual dance five more times.

Either way, no merge here until the full matrix — `sdk`, `providers (aws)`, `providers (macos-dev)`, and `aws-terraform-validate` — has actually run against this content. `aws-terraform-validate` is the one that matters, for exactly the reason the review gives: this PR changes emitted Terraform, and structural reads are not `terraform validate`.

## Non-blocking note — quiz line — fixed in `836e512`

`www-dev/src/pages/learn/env-and-secrets.astro:124` now matches the other three spots in the file:

```diff
-explanation: "Correct! A cryptographically random string is generated for you at deploy time — no manual setup required."
+explanation: "Correct! 32 cryptographically random bytes, hex-encoded (64 characters), are generated for you at deploy time — no manual setup required."
```

## Non-blocking note — the missing D-45 record

Agreed, and it is not this PR's to fix. `git grep D-45` is empty; the reservation for #171 exists only in PR prose. It is introduced by #176 and becomes permanent the moment this lands on top of it — so it should land **on #176**, before that merges, not as a follow-up afterwards. Please file it there.

## Verification

All OBSERVED, run against this head (`836e512`) in the branch worktree. Prose-only change, so the expectation was "unchanged", and that is what happened:

| Suite | Result |
|---|---|
| `sdk` | **280 pass, 0 fail** (after `bun run build` — the CLI suite needs `dist/cli.js`) |
| `providers/macos-dev` | **102 pass, 0 fail** |
| `git status` | clean; branch in sync with `origin/chore/gov-secret-generator` |

Two commits, split by top-level directory per the repo's subtree-split convention: `spec:` and `www-dev:`.

ziadsawalha:
Correcting myself: **you're right that #195 is not a second route for this PR, and my earlier reply was wrong to present it as one.**

I re-derived it rather than just conceding. All three refs this PR's event resolves against still carry the old filter:

```
$ for r in origin/main origin/chore/gov-resource-registry origin/chore/gov-secret-generator; do
    git show "$r:.github/workflows/ci.yml" | sed -n '3,5p'
  done
on:  pull_request:    branches: [main]
on:  pull_request:    branches: [main]
on:  pull_request:    branches: [main]
```

#195's change exists only on `ci/run-on-every-pull-request`. Merging it into `main` changes none of those three refs, so this PR stays at zero checks. And once #176 merges and this retargets, the base *is* `main`, which the existing filter already matches — so #195 is not required for this PR either. Orthogonal in both directions, exactly as you put it.

What I got right by accident was the conclusion for the *rest* of the stack, and only because #195's content reaches those branches through the train merges — not because merging it to `main` does anything on its own. That distinction is the part I skipped, and it is the part that determines sequencing.

So the path here is the one you specified originally, unchanged: **#176 merges → this retargets to `main` → push a sync commit.** The retarget alone will not do it — with no `types:` key the workflow fires on `opened`, `synchronize` and `reopened`, and a base change is none of the three. #195's own file comment makes the same point.

#176 is ready for that first step: 12/12 green, `mergeStateStatus: CLEAN`, and your APPROVE stands. The D-45 stub landed there in `62bc959`, so the numbering hole closes before it merges rather than after.

On your `TERRAFORM_BIN=tofu` run — that is a genuinely useful narrowing and worth keeping on the record: 13/13 valid, 4 of them declaring `generator: secret`, so `hashicorp/random ~> 3.6` really did initialize and `random_bytes { length = 32 }` really did schema-validate. Agreed it is a signal and not the recorded gate; I am not treating it as discharging the blocker.

Two things I'll fix rather than argue:

- **The PR description still carries the corrected-away claim.** `DESIGN.md` is fixed; the description isn't. Editing it now.
- **The 2-of-23 drafts nuance.** You flagged that "sit outside that evidence" understates, since 2 of the 23 do carry `health_check_passed: true`. It errs safe, but the sentence is meant to be exactly true rather than conservatively true — I'll tighten it in the same push.

ziadsawalha:
Both done.

**Drafts count — `c6bcd6a`.** Verified your 2-of-23 before changing anything:

```
$ for f in $(git grep -l 'generator: secret' -- catalog/drafts/); do
    grep -l 'health_check_passed: true' $(dirname $f)/metadata.yaml
  done
catalog/drafts/anythingllm/metadata.yaml
catalog/drafts/n8n/metadata.yaml
```

`spec/DESIGN.md:517` now reads *"a further 23 `catalog/drafts/` entries declare it, of which 2 (`anythingllm`, `n8n`) are harness-tested green and the remaining 21 sit outside that evidence."*

Taking the correction even though it errs safe: the sentence is meant to be exactly true, not conservatively true. "Understates" and "overstates" are the same defect once someone re-runs the count.

**PR description** — updated in place, no longer carries "all **26**… The whole catalog has already been validated against this output." It now matches `DESIGN.md`.

Head is `c6bcd6a`. Blocker 2 is untouched and still stands; nothing here claims CI coverage.

ziadsawalha:
CI is green on `1670dd8`. The blocker is discharged.

Your review landed at 23:15:26Z. `git show -s --format=%cI 1670dd8` → **23:15:27Z** — the sync commit was made one second later, and the run was created at 23:16:01Z. The two crossed in flight, which is why the review couldn't see it.

`gh api .../actions/runs?branch=chore/gov-secret-generator` → **`total_count: 1`**, event `pull_request`, `head_sha` `1670dd8`, conclusion `success`: [run 32604751596](https://github.com/launchfile/launchfile/actions/runs/32604751596).

The four you named:

| Check | Result | Duration |
|---|---|---|
| `SDK` | pass | 15s |
| `Provider (aws)` | pass | 11s |
| `Provider (macos-dev)` | pass | 11s |
| `AWS provider — terraform validate` | pass | 1m54s |

`gh pr checks 177 --json name,state` → **12 checks, 0 not-SUCCESS**. The other eight: `CLI`, `Provider (docker)`, `Website (www-dev)`, `Website (www-org)`, `Workflow Lint`, `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)`.

**Which trigger fired — derived, not read off the API.** The runs API doesn't expose the `pull_request` action, so I bounded it instead: `issues/177/timeline` contains no `reopened`, `closed`, or `head_ref_force_pushed` event, ever — the only structural event on this PR is `automatic_base_change_succeeded` at 22:42:51Z. `ci.yml` declares no `types:`, so the default set is `opened`/`synchronize`/`reopened`; `opened` and `reopened` are excluded by that timeline, leaving `synchronize`. Your sync-push route is what worked. The reopen fallback was never needed.

**The terraform gate is now the gate, not an approximation.** Your 13/13 under OpenTofu was strong evidence; `hashicorp/setup-terraform` running against the same emitted HCL is the recorded check. Confirmed at head that it's the surface this change introduces: `providers/aws/src/translate.ts:364` emits `resource "random_bytes" { length = 32 }` and `:386` pins `hashicorp/random` at `~> 3.6`.

Re-verified locally at the same head, alongside CI rather than instead of it:

```
sdk                        289 pass / 0 fail
providers/aws               40 pass / 0 fail
providers/macos-dev        109 pass / 0 fail
conformance.ts --check      CONFORMANCE.md is up to date
TERRAFORM_BIN=tofu validate:tf --examples    13/13 valid
```

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. `origin/main` is still `1320b72` and `git rev-list --count HEAD..origin/main` → **0**, unchanged since the run — so `main` is fully merged into the branch, and the merge ref CI resolves therefore has the same tree as the head it reported against.

Nothing else outstanding from the three reviews: blocker 1 closed at `6a9560a`, the drafts count tightened at `c6bcd6a`, the description corrected to 25-of-26 with the 2/21 drafts split, and the D-45 hole discharged by #176's `62bc959` rather than deferred.

Ready for re-review.

PR #52, opened 2026-06-09
feat: install / dev — source-mode lifecycle commands (D-36, supersedes #45)

ziadsawalha:
Implements **#79** (`install` / `dev` source-mode lifecycle commands), resting on the **#77** execution-mode taxonomy — both ratified as Author. **Supersedes #45.**

Reworked from the original #45 design (compound `dev:<stage>`, D-35) to the ratified flat model.

## What this adds (D-36)

- **`install` / `dev`** — two well-known `commands:` keys (source prepare / source run), open-record, zero schema change.
- **`source:`** — optional component working-directory field (defaults `build.context` → repo root).
- **Resolution** — run precedence `dev` > `image` > `start`; prepare = `install ?? build`, on demand; only `prepare`/`run` are mode-aware (`release`/`bootstrap`/`seed`/`test` stay mode-invariant — no `dev:<stage>`).
- **`launchfile dev`** verb — source-mode launch (forces the native provider).
- **Source-mode guard** — `dev` / `up --native` on an image-only app fails fast with a pointer to `launchfile up`, instead of a silent no-op.

## Decision number

Lands as **D-36** — which also resolves the number collision the original branch had with `$app.authority` (the merged D-35). #86 (three-homes litmus) takes D-37. The stale `dev`/D-35 entry the earlier review flagged was a **gitignored build artifact**, never committed source; this PR adds the correct flat keys to the committed schema/SDK.

## Scope

Spec + SDK + providers + CLI only. **Catalog source-mode entries** (concentrator, and exemplars) are intentionally **split to a follow-up** — `catalog/apps/` is "tested and verified," and source-mode catalog entries can't be validated without each app's source tree.

## Verification

- Tests: SDK **191** · macos-dev **64** · docker **79** · CLI **34** — all green.
- All 10 spec examples `validate`.
- Live `launch dev` on a throwaway source app: `install` ran in the `source:` dir, `dev` started, `$PORT` injected, `/health` → `ok`, `down` clean.
- `launch up` (docker artifact) unaffected (ghost compose generates cleanly).

PR #193, opened 2026-08-22
spec: per-stage failure semantics, one duration grammar, command interpretation (D-48)

### LINKED ISSUE #180: RFC: per-stage failure semantics and one duration grammar (zero new fields)
## RFC: per-stage failure semantics and one duration grammar — zero new fields

Two ratifications over the shape the spec already has. **No new fields, no schema change, no syntax.**

### Problem 1 — the per-stage dispositions are mostly unstated, and one provider ignores a rule that already exists

**Corrected after the steward review**, which caught this section overstating its case. The spec is **not** silent about `release`: `SPEC.md` already says it *"runs before `start` in an ephemeral container and fails the whole deploy on error"*, and `PROVIDERS.md` §3 already lists the release slot as running per deploy.

So `@launchfile/docker` is **already non-conformant against the shipped spec**, and fixing it ratifies nothing — that half is a provider bug, tracked separately. What is genuinely unstated is every *other* stage's disposition, and the resources-ready side of the release window.

The spec never says which of the remaining lifecycle stages fail a deploy and which report and continue. The two reference providers reached opposite answers for the same file:

| | `@launchfile/macos-dev` | `@launchfile/docker` | `@launchfile/aws` |
|---|---|---|---|
| `commands.release` | **runs it**, and a non-zero exit fails the launch (`src/provider.ts:320-331`) | **never runs it** — `grep -c 'release' compose-generator.ts` → **0** | **runs it** (`src/translate.ts:141-143`) |

The divergence is **1-of-3, not 2-of-2** — corrected after review. aws already honors `release`, which makes Docker the clear outlier rather than an even split.

So a Launchfile declaring a database migration in `release:` gets its migration on one provider and silently skips it on the other, then starts the app against an un-migrated database. `catalog/drafts/chatwoot` and `catalog/drafts/hedgedoc-v2` both declare `release`.

That is `P-5` failing in this repo's own reference implementations.

### Problem 2 — one grammar, two parsers, and a silent default

`commands.*.timeout` and `health.interval` / `timeout` / `start_period` all take durations. The spec never gave the grammar. Both providers shipped **byte-identical** parsers that are more permissive than any documented rule, and both swallow a parse failure:

```ts
// providers/docker/src/bootstrap.ts:61-63  AND
// providers/macos-dev/src/bootstrap.ts:89-91  — identical
export function parseDuration(s: string): number {
  const match = /^(\d+)\s*(ms|s|m|h)$/.exec(s.trim());
  if (!match) return 120_000;
```

Write `timeout: "5 minutes"` and you do not get an error. You get **two minutes**, silently. The author's declared intent is discarded and replaced with a number they never wrote.

### Corrections and additions after the first review

The steward returned REJECT with four blockers. All four are addressed; the substantive changes to the proposal:

1. **The failure table is now keyed on *slot*, not command name.** Keying on the name overlapped: D-38 resolves source prepare as `install ?? build` and source run as `dev ?? start`, so `build` and `start` each matched two rows. `catalog/drafts/hedgedoc-v2` — `build:` + `start:` + `release:`, no image/dev/install — is exactly that shape, and is one of the two files cited as motivation. The prepare and run rows now fail **the invocation**, which is what makes them disjoint.
2. **A third duration parser existed and is now conformed.** `providers/macos-dev/src/health.ts` had its own `^(\d+)(ms|s|m)$` with a silent `return 0` — no hour unit. `interval: "1h"` was spec-valid, passed `validate`, worked on Docker, and became `0` there, making the health check unpassable. Health is where **11 of 11** catalog duration declarations live; `commands.*.timeout` is used by **0 of 113** files, so the branch had conformed the parser nothing uses.
3. **Command interpretation is now stated.** It was ambiguous and the providers diverged — docker splits on whitespace and execs `argv[0]` with `shell: false`, macos-dev runs a POSIX shell. The spec now says shell, because the catalog already depends on it: `paperclip`'s bootstrap is a multi-statement script, and the argv split makes docker try to exec a binary named `CFG="$PAPERCLIP_HOME/..."`. That is a live bug, filed as **#185**.
4. **The impact numbers were wrong** and are corrected below.

### Proposed change

**(1) A complete, non-overlapping per-stage failure table** in `SPEC.md`:

- `build`, `release`, `start` — **fail the deploy**
- `install`, `dev` — fail the invoking session or workflow; source mode (D-38), never a deployed-state concern
- `bootstrap`, `seed`, `test`, custom commands — **on demand**: failure is reported to the invoker and never affects deploy status

Timeout expiry is a failure with the same disposition as any other failure of its stage.

One clause is **new normative precision, not codified prose**: `release` *"runs after the component's required resources are provisioned and ready, and before `start`"*. The spec previously pinned only the `start` side of that window; the resources-ready side is added here.

**(2) One duration grammar**: `^(\d+)(ms|s|m|h)$`, governing every duration-valued field.

This **tightens** rather than codifies — both shipped parsers accept surrounding and internal whitespace, and the ratified grammar accepts neither. An unparseable duration becomes a `validate`-surfaced **non-fatal warning** (D-40's validate-only diagnostic precedent), and a provider **MUST NOT** silently substitute a default — it surfaces the error.

**(3) Provider conformance** (`PROVIDERS.md` §10.10): a provider executing a deploy MUST run a declared `release` after resources are ready and before `start`, and MUST fail the deploy on its error. Numeric timeout *defaults* stay provider-side, documented by each provider.

### Motivation

- **2 catalog entries declare `release`** (`chatwoot`, `hedgedoc-v2`) and get opposite behavior across the two providers today. Both are in `drafts/`, which I want to state plainly rather than round up — but the divergence is in the shipped providers, not in the drafts.
- **Duration usage, measured** (corrected): **11 of 113** files declare any duration, all of them health durations; **0 of 113** declare `commands.*.timeout`; and all 11 parse correctly today, so **zero catalog files are currently mis-parsed**. The silent default is a latent trap, not a live breakage — except through the health parser in point 2 above, which *is* live for `1h`.
- **Both defects are in this repo's own reference implementations**, which is what makes this a `P-5` argument rather than a hypothetical.

### Principle self-assessment

| Principle | Assessment |
|---|---|
| **P-5** provider-translatable | **The point.** One file must not produce a migration on one provider and skip it on another. |
| **P-6** it's just YAML | **Pass.** Zero schema change, zero new syntax. |
| **P-11** intent vs execution | **Respected by construction.** The file gains nothing. The spec binds stage *meaning*; the provider keeps mechanism — how it aborts, how it surfaces, and its own default budgets. |
| **P-4** human-writable | **Pass.** Authors write nothing new. |
| **P-9** unambiguous | **Pass.** One grammar everywhere a duration appears. |
| **P-13** additive extensibility | **Pass.** Warning-only enforcement; no field changes meaning; every existing file stays valid. |
| **D-37 / D-38** | The failure table maps onto the existing slot taxonomy: prepare/release/run slot failures fail the launch; on-demand slot failures are reported. |
| **D-40** | Uses its validate-only diagnostic precedent for the duration warning. |

**Risks, stated plainly:**

1. **The grammar tightening is a real behavior change for providers**, not a no-op. A file using `timeout: "30 s"` parses today and warns after. No catalog file uses a whitespace form — checked — and the enforcement surface is warning-only, so no file becomes invalid. But a provider that adopts the strict grammar in its *parser* will now throw where it previously returned 120s, and that is a deliberate hard-fail replacing a silent one.
2. **Docker gains a stage it never ran.** Making the Docker provider execute `release` is new behavior for anyone relying, knowingly or not, on it being skipped. This is the bug being ratified away, but it will change what their deploys do.
3. **Reversibility: high for the format** (nothing to reverse), **moderate for the provider contract** — once providers hard-fail on bad durations, relaxing again is another conformance change.

### Open question for review

Should the **`validate` warning** for an unparseable duration be an **error** instead? The argument for warning is D-40's precedent and P-13. The argument for error is that unlike a property typo, a bad duration has a *defined* wrong outcome today (120 seconds), and the whole point of this decision is that silently substituting a number the author did not write is unacceptable — a warning still lets it through.

ziadsawalha:
Ratifies four things about the lifecycle, over the shape the format already has. **No new fields, no schema change, no syntax.** Records them as **D-48**.

Closes #180

## What it settles

**(1) A per-slot failure table.** Keyed on the `PROVIDERS.md` §3 *slot*, not the command name — because D-38 resolves source prepare as `install ?? build` and source run as `dev ?? start`, so `build` and `start` each appear in two modes. Prepare and run **fail the invocation**: the deploy when deploying, the session when running from source. That is what makes the table non-overlapping.

**(2) One duration grammar**, `^(\d+)(ms|s|m|h)$`, for every duration-valued field. An unparseable value is a `validate` warning (D-40's precedent), and a provider **MUST NOT** silently substitute a default.

**(3) Provider conformance** — a declared `release` runs after resources are ready and before the run slot, and fails the deploy on error.

**(4) Command interpretation** — a command string is interpreted by a **POSIX shell**.

## The defects this closes

Both were shipped in this repo's own reference providers:

- **Docker never ran `release`.** `grep -c release providers/docker/src` on `main` → **0**, while macos-dev runs it and fails the launch. Same file, opposite outcome. (aws already honored it — so Docker was the outlier, not an even split.)
- **A silent 120-second default.** Both bootstrap parsers shipped `if (!match) return 120_000`, byte-identical. Write `timeout: "5 minutes"` and you got two minutes, silently.

Plus one the review found: **a third duration parser** in `providers/macos-dev/src/health.ts` with no `h` unit and a silent `return 0`. So `interval: "1h"` was spec-valid, passed `validate`, worked on Docker, and here became `0` — making the health check unpassable. Health is where **11 of 11** catalog duration declarations live; `commands.*.timeout` is used by **0 of 113** files.

## How the #180 verdict was addressed

Two `REJECT` rounds, seven blockers, then `ACCEPT`:

| Blocker | Resolution |
|---|---|
| Table overlapped under D-38 source mode | Re-keyed on slot. Verified: all 113 catalog files, five distinct component shapes, each lands in exactly one row |
| Third duration parser un-conformed | Routed through `parseDurationMs`, fenced with tests |
| Problem-1 premise false | `SPEC.md` already said `release` fails the deploy, so Docker was already non-conformant and its fix ratifies nothing. Corrected in the issue and in D-48's *Why* |
| `release` argv/shell divergence | Docker now uses `sh -c` as a single argv element |
| § Command interpretation had no `D-*` and no conformance hook | Ratification (4) in D-48 with the security trade-off recorded; `PROVIDERS.md` §10 item 11 |
| `www-dev` published the superseded table | Re-keyed to match SPEC row-for-row; command interpretation taught; site builds |
| §10.10 dangling health reference | § Failure semantics now gives health a disposition |

**On the shell change**, since it reverses a documented posture: both providers argv-split *deliberately*, to avoid shell-injection exposure. That is overruled here with the reason on the record — the command is the app's own Launchfile content the provider is already about to execute, so a shell grants no privilege the caller lacked, while the split silently truncated documented usage. `providers/docker/src/shell.ts` uses `execFile`, so the command crosses as data in one argv element and is interpreted only by `sh` *inside the container*.

## Known gap, recorded

`bootstrap` still argv-splits on both providers — a live violation of the §10.11 this PR introduces. **Pre-existing**, tracked in **#185**. Found while settling blocker 4: `paperclip`'s bootstrap is a multi-statement script, and the split makes Docker try to exec a binary named `CFG="$PAPERCLIP_HOME/..."`.

## Tests

sdk 290/290 · docker 103/103 · macos-dev 105/105 · aws 40/40 · CLI 42/42 · all typechecks clean · 113 catalog files, 0 new warnings

## Stacking

Based on `chore/gov-secret-generator` (**D-47**). D-45 is deferred (#171) and its number stays reserved.


ziadsawalha:
All three blockers addressed, plus one defect that surfaced while verifying — and it would have broken CI.

## Blocker 1 — secret disclosure — `ec3dfaa`

Took your route (b): a `redact.ts` mirroring `@launchfile/macos-dev`, rather than carrying the unresolved command. Route (a) only covers the echo at `:165`; it does nothing for `:185-186`, which print the **child's** stdout/stderr — and a migration tool echoing its own connection string is the more likely leak in practice. (b) also closes the pre-existing `bootstrap.ts:250` echo you flagged.

Two layers, same API as the sibling provider:

1. **A registry of exact secret values.** Generated passwords/secrets (`compose-generator.ts:49,57`), persisted `state.secrets` (`state.ts:73`), and every `secrets` map handed to the release planner (`release.ts:100`) register at creation/load time. Values under 8 characters are not registered — short strings collide with unrelated text, and scrubbing them corrupts the output the redaction exists to protect.
2. **A pattern scrub for `scheme://user:pass@host`**, which catches credentials that never passed through this provider — a connection string written literally in an `env:` value and interpolated into a release command.

Verified live, not just by the tests:

```
raw command      : seed-admin --token S3CR3T-LIVE-VALUE-abcdef123456
echoed (redacted): seed-admin --token [REDACTED]
stdout (redacted): connecting with token [REDACTED]
stderr (redacted): auth failed for [REDACTED]
url-embedded     : postgres://admin:[REDACTED]@db:5432/app
```

All three sites you named (`:165`, `:185`, `:186`) plus the URL pattern. 210 lines of tests in `__tests__/redact.test.ts`.

Your scope framing is right and I am not inflating it: **live exposure today is zero** — neither catalog file declaring `release` uses `$secrets.*`. This is a latent hazard in new code.

## Blocker 2 — `spec/DESIGN.md` re-keyed — `e213091`

D-48's ratification (1) no longer keys the failure rule on command names. It keys on slot, matching the SPEC table for source mode — the framing #180's blocker 1 asked for.

## Blocker 3 — CI — stands, and I am claiming nothing

```
$ gh pr checks 193
no checks reported on the 'chore/gov-failure-semantics' branch
```

Base is `chore/gov-secret-generator`. Clears when the train lands, this retargets to `main`, and a sync push follows. Noted that #177 has advanced to `c6bcd6a` while this branch is based on `ec18fcb`; I have not rebased — the operator sequences that.

## The thing blocker 3 was hiding — `c5d6f83`

I added the wiring test you asked for (`provider.ts:265-284` had none) and it **passed locally and would have failed CI**:

```
bun test      -> 118 pass, 1 fail, 2 errors
                 TypeError: importOriginal is not a function
bun run test  -> 120 passed
```

`ci.yml`'s providers matrix runs **`bun test`**, not `bun run test`. The package's `test` script is `vitest run`, and `vi.mock("…", async (importOriginal) => …)` is a vitest-only API — bun's `vi.mock` passes no such argument, so the factory throws before a single test executes. Both factories now list the exports the code under test pulls, which needs no runner-specific API. Both runners agree:

```
bun test      -> 120 pass, 0 fail
bun run test  -> 120 passed
```

That file was the only `importOriginal` use in the repo, so this is isolated — but it is worth naming the general hazard: **`bun test` and `bun run test` are different runners here, and CI uses the one local development doesn't.** A green local suite is not evidence about CI for this repo. It only stayed invisible because no CI has ever run on this PR, which is precisely your blocker 3 doing damage rather than just sitting there.

## Verification — OBSERVED, under CI's runner (`bun test`), build order sdk → providers → CLI

| Suite | Result |
|---|---|
| `providers/docker` | **120 pass, 0 fail** (was 103; +17 from `redact.test.ts` and the wiring test) |
| `providers/macos-dev` | **105 pass, 0 fail** |
| `packages/launchfile` | **42 pass, 0 fail** |
| `tsc --noEmit` (docker) | clean |

Head is `c5d6f83`.

ziadsawalha:
Both blockers closed, and CI — which finally ran — caught a third thing that came in with the redaction fix.

## Blocker 1 residual — the pino sink — `9f6c43c`

You were right, and the reproduction is exact. `ec3dfaa` scrubbed the console echoes and the child's streams and left `shell.ts`'s four `log.debug` calls recording `args` raw. Confirmed both directions with **only `LAUNCHFILE_LOG_DIR` set** — no debug flag, exactly as you described, because `logger.ts` pins the root level to `trace` whenever a log dir exists:

```
at c5d6f83 (before)
  "args":["seed-admin","--token","S3CR3T-LIVE-VALUE-abcdef123456"]
at 9f6c43c (after)
  "args":["seed-admin","--token","[REDACTED]"]
```

All four sites (`:33`, `:56`, `:92`, `:114` pre-patch) now take a `safeArgs = args.map(redactSecrets)` computed once per function, alongside the existing `display`. Your diagnosis of why `REDACT_PATHS` can't cover it is right and I put it in the comment: fast-redact matches key paths, and the value is a string inside an array.

Your point about *why* mirroring macos-dev didn't surface this — its `shell.ts` has no logger — is the useful part. The mirror was faithful; the sink was local to this provider.

## Blocker 2 residual — the docs takeaway — `08b4aff`

Took your suggestion verbatim. The bullet was name-keyed and contradicted the slot table this same PR adds 190 lines above. `grep -rn "build/release/start fail\|install/dev fail" www-dev/src/` now returns nothing.

## New, from CI — a high-severity ReDoS in `redact.ts` — `3375a0f`

CI ran on this PR for the first time and **CodeQL failed**: `js/polynomial-redos`, high, at `redact.ts:69`. The credential-URL pattern opened `[a-zA-Z][a-zA-Z0-9+.-]*:\/\/`. Unbounded, every starting offset in a run of scheme-legal characters rescans that whole run looking for `://` — quadratic. A denial of service in the redactor whose job is to protect the log.

Measured, same input, old pattern vs new:

```
n=10000    54ms ->  0ms
n=40000   804ms ->  1ms
n=80000  3293ms ->  3ms
```

Fixed by bounding the scheme repetition to `{0,31}`. RFC 3986 schemes are `ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )` and the longest IANA-registered one is far under 32, so nothing real is excluded — `postgres://`, `https://` and `amqp+ssl://` all still redact, verified.

**`providers/macos-dev/src/redact.ts` carries the identical pattern.** It is not in this PR's diff, predates it on `main`, and has never been flagged — CodeQL analyses the PR diff, so it was simply never scanned. Out of scope here; it wants a follow-up, and I'd rather you filed it than have me widen this PR.

One correction to something implied earlier: the two `redact.ts` files are **not** byte-identical — `diff` on them at this head reports differences. "Mirrors macos-dev" is accurate about API and structure, not literally byte-for-byte, and this ReDoS fix widens that gap by one line.

## Verification — OBSERVED, CI's runner

| Check | Result |
|---|---|
| `providers/docker` `bun test` | **120 pass, 0 fail** |
| `providers/docker` `bun run test` (vitest) | **120 passed** |
| `providers/docker` `tsc --noEmit` | clean |
| `www-dev` `tsc --noEmit` | clean |

Thank you for the repo-wide runner sweep — `vi.hoisted`, `stubEnv/stubGlobal`, `doMock/unmock`, `importActual/importMock`, `expect.soft/poll`, `vi.waitFor`, `onTestFinished`, `describe.concurrent`, fake timers, snapshots, all zero hits. That is a more definitive answer than I had, and your note that the factory is evaluated lazily — so a probe that doesn't import the mocked module proves nothing — is the part that would have caught me out.

## Blocker 3 — CI

Now partly answered by events: this PR **did** get a full run (12 of 13 green, `terraform validate` still going, CodeQL red until `3375a0f`). #176 merged, #177 synced to `main` and went 12/12 green, so the train is moving. The `aws-terraform-validate` gate on *this* content is the piece still outstanding.

Head is `3375a0f`.

ziadsawalha:
Comment corrected in `323a96e`. You were right and the error was mine.

`microsoft.windows.camera.multipicker` is **36** characters — verified — and RFC 3986 sets no length ceiling at all, so "the longest IANA-registered one is well under this bound" was simply false. Worse, it was load-bearing in the wrong direction: a future reader taking that justification at face value would "fix" a long-scheme miss by raising the bound, restoring the quadratic scan for nothing.

The comment now states the reason you identified — the pattern is unanchored, so a longer scheme just moves the match start:

```
aaaa…(40 chars)://user:s3cr3t@host              -> …://user:[REDACTED]@host
microsoft.windows.camera.multipicker://u:pw@host -> …multipicker://u:[REDACTED]@host
postgres://admin:hunter2@db:5432/app             -> postgres://admin:[REDACTED]@db:5432/app
```

Code unchanged; `bun test` 120/120, typecheck clean.

That is a better safety property than I claimed, and I'd rather the comment said so than kept a justification that happens to reach the right conclusion from a wrong premise.

## On the `shell.ts:69` error-serializer hypothesis

Agreed with how you handled it — reported as a hypothesis with a fix, not as a finding, because reachability wasn't demonstrated. For the record I agree it isn't this PR's: the shape is byte-identical on `main`, the release path passes `allowFailure: true` so the `result`-bearing error is never constructed there, and your four probes against a real compose binary came back negative.

It is still worth closing eventually — pino's default `err` serializer walking own enumerable properties is exactly the kind of thing that becomes reachable the moment someone adds a `withSpan` around a secret-bearing call. A `serializers.err` running `result` through `redactSecrets` is cheap insurance. Filed-or-not is your call; I haven't opened one, so as not to duplicate whatever you decide alongside #203.

## #203 and alert #52

Both dispositions look right. On #203 in particular — "zero alerts on `main` is a coverage artifact, not a clean bill of health" is the part worth carrying forward, because it applies to every file in the repo that has never appeared in a PR diff, not just that one.

Head is `323a96e`.

Issue #191, opened 2026-08-22
RFC: a provider MUST NOT fabricate a value for an unsupplied required env var


ziadsawalha:
## RFC: a provider MUST NOT fabricate a value for an unsupplied `required` env var

One conformance rule, scoped by verb. **No new fields, no schema change.**

### Problem

`env.<NAME>: { required: true }` is the app's own statement that **it cannot start without this value**. When nobody supplies one, the reference Docker provider invents one — by guessing from the variable's *name* (`compose-generator.ts:728-737`):

```ts
if (envVar.required) {
  const lowerKey = key?.toLowerCase() ?? "";
  if (lowerKey.includes("url") || lowerKey.includes("domain") || lowerKey.includes("origin"))
    return "http://localhost";
  if (lowerKey.includes("email") || lowerKey.includes("mail"))
    return "test@localhost";
  return "PLACEHOLDER";
}
```

So `outline`'s `URL` — declared `required: true` with no default — becomes `http://localhost`. Outline then builds every share link, invite and callback URL against that value. The deploy reports success. The app runs. Everything it emits points at a host nobody else can reach.

`posthog` and `flowise` are in the same position.

**This is worse than a warning-then-failure.** A warning that a required value is missing, followed by a failed start, tells the operator exactly what to fix. Substituting a value converts a stated precondition into a **silent misconfiguration** that surfaces later, somewhere else, as behaviour nobody connects back to the missing variable. `PLACEHOLDER` as a database password fails loudly; `http://localhost` as a public URL does not.

### Proposed change

A provider **MUST NOT** substitute a fabricated value for an unsupplied `required` env var.

**Fabricated** means a value the provider invented rather than obtained — a constant like `PLACEHOLDER`, or a guess derived from the variable's name.

**Unsupplied** means none of the file's own sources yields one: no `generator:`, no `default:`, no `set_env` binding, and nothing from the provider's own operator-facing channel.

The obligation is **keyed to the verb**, because "fail" is not available everywhere:

| Verb | Obligation |
|---|---|
| **Deploying** (`up`, and anything that provisions or runs) | Obtain the value from the provider's own operator channel, or **fail with a clear message naming the variable**. Never invent one. |
| **`translate`** — no launch at which to fail | Leave the variable **absent** from the emitted artifact and record it as an unmet requirement. |
| **Inspection** (`env`, `status`, `list`, `logs`) | Report the variable as unsupplied rather than omitting it or showing an invented value. |

[...]

PR #194, opened 2026-08-22
spec: env-value provenance — four declaration classes with per-layer obligations (D-49)

### LINKED ISSUE #181: RFC: env-value provenance — four declaration classes with per-layer obligations
## RFC: env-value provenance — four declaration classes with per-layer obligations

Names the four ways an env value can come to exist, and states what each layer — author, provider, operator — owes for each.

**Scope: spec + docs + one provider fix.** No schema change, no new field. An earlier version of this issue said "no code" — that was wrong, and the reason is the strongest argument for the decision:

**Two reference providers already disagree, today, in tree.** `providers/docker/src/compose-generator.ts:709` and `providers/aws/src/translate.ts:953` both resolve `generator` → `default` → `required`. `providers/macos-dev/src/env-writer.ts:163` fills the `default` first, and `:208` then skips any key already set — so **`default` wins over `generator` there**. Observed, not inferred:

```
{ FOO: { default: "author-literal", generator: secret } }

macos-dev  FOO = author-literal
docker     FOO = 2b8109bb591bf75d6eda311c839ec480648c4a43bca5d1b6eb0af3bc7f9f0544
```

Same declaration, two values. That is a live `P-5` violation, and ratifying `generator:` first makes macos-dev non-conforming on merge — so the fix ships with the decision.

### Problem

`env:` already carries `generator:`, `default:`, `required:` and the bare declaration. What the spec never says is **who is responsible for the value** in each case, so the same declaration gets different treatment from different tools, and an author has no way to know which one they are asking for.

The gap is most visible where the fields combine. `catalog/apps/miniflux/Launchfile:27-31`:

```yaml
ADMIN_PASSWORD:
  required: true
  sensitive: true
  generator: secret
```

Is that "the operator must supply it" (`required: true`) or "the platform mints it" (`generator:`)? Both fields are present and the spec does not say which wins. A tool that reads `required` first will prompt a human for a password that the platform was about to generate.

### Proposed change

Every declared env value falls into **exactly one** of four provenance classes, decided by an explicit precedence over the fields that already exist — **`generator:` → expression `default:` → literal `default:` → neither**:

| # | Class | Trigger | Who owes the value |
|---|---|---|---|
| 1 | **Minted** | `generator:` present | The **platform** generates once, then preserves. A `generator:` **satisfies** `required:`. |
| 2 | **Derived** | `default:` is an expression over platform-resolved inputs (`$app.*`, resource properties, `$secrets.*`) | The **provider**, at resolution time. |
| 3 | **Author default** | a literal `default:` | The **file author** chose a starting value. Nobody needs to supply anything. |
| 4 | **User-supplied** | no generator, no default | Only a **person** can supply it. Covers `required: true` (must supply before start) and the bare declaration (may supply). |

The precedence is the load-bearing part: it resolves the miniflux case as **minted**, so `required: true` is satisfied by the generator and no human is prompted.

### Conformance at adoption

Stating this here so the RFC of record carries it, not only the ratified text: the **minted** obligation — generate once, then preserve — is met today only for the top-level `secrets:` block. It is **not met for `env:`-level `generator:` values**, which both reference providers re-mint on every `up`. That is 26 declarations across 23 catalog apps, including this decision's own examples. The obligation is stated because it is right; the gap is recorded because it is real, and tracked in **#186**.

### Two further rules this decision states

Both are in the ratified text and were missing from an earlier version of this issue. Flagging them so they are reviewed rather than inherited:

1. **A supplied `default:` satisfies `required:`.** This changes what `required: true` means alongside a default — from "a person must supply this" to "this must be non-empty at start". It is a second precedence decision, arguably larger than the generator one.
2. **A literal `set_env` value is deliberately constant.** An expression wiring is derived; a literal wiring is passed through verbatim, because overriding it breaks the resource wiring the author declared. So a literal in `set_env` and an identical literal in `default:` classify differently — the class depends on shape **and** location, which the "precedence over existing fields" framing above does not by itself convey.

### Motivation — all four classes exist in the catalog today

Classified all 72 shipped apps by the proposed precedence:

```
  minted            26 declarations / 23 apps
  derived           23 declarations / 20 apps
  author-default    34 declarations / 22 apps
  user-supplied     10 declarations /  5 apps
  TOTAL             93
```

**Corrected after the steward review**, which reproduced 93 where this issue originally said 86. Two mistakes in the first pass: it skipped component-level `components.*.env:` blocks, and it did not treat a **scalar shorthand** (`FOO: "$app.url"`) as a `default:` per `SPEC.md:404`. The three `APP_KEY: "base64:${secrets.app-key|base64}"` declarations in `snipe-it`, `monica` and `firefly-iii` are scalar shorthands and classify as **derived** — the same three files this decision cites as its composition example, so the original count contradicted its own worked example.

The classifier, so the numbers are reproducible rather than asserted:

```ts
const classify = (v) => {
  if (v == null) return "user-supplied";                                  // bare declaration
  if (typeof v !== "object")                                             // scalar shorthand = default
    return String(v).includes("$") ? "derived" : "author-default";
  if (v.generator) return "minted";
  if (typeof v.default === "string" && v.default.includes("$")) return "derived";
  if ("default" in v && v.default !== undefined) return "author-default";
  return "user-supplied";
};
// over doc.env plus every doc.components.*.env, across catalog/apps/*/Launchfile
```

The user-supplied row merges `required: true` (4: `flowise` ×2, `outline`, `posthog`) and bare declarations (6: `paperclip` ×2, `remote-claude-concentrator` ×4), which the decision keeps distinct — the obligation is identical, only the timing differs.

Worked examples, each verified in the shipped file:

- **Minted** — `miniflux` `ADMIN_PASSWORD` (`required` + `generator`, the ambiguous case above)
- **Author default** — `snipe-it` `APP_URL: default: "http://localhost"` (`catalog/apps/snipe-it/Launchfile:26-27`)
- **User-supplied** — `outline` `URL: required: true` with no default (`catalog/apps/outline/Launchfile:31-33`)

So this is not a new taxonomy imposed on the format. It is a name for a structure **93 declarations across the catalog already have**, plus one precedence rule to settle the case where the fields overlap.

### Principle self-assessment

| Principle | Assessment |
|---|---|
| **P-1** app-focused | **Pass.** Provenance is a property of the app's own declaration, not of any deployment target. |
| **P-3** machine-generatable | **Strengthened.** A generator can now target a defined class instead of guessing from field combinations. |
| **P-4** human-writable | **Strengthened.** An author can tell what they are asking for, and the miniflux shape stops being ambiguous. |
| **P-5** provider-translatable | **The point.** Two providers reading the same declaration currently owe different things; this fixes that. |
| **P-6** it's just YAML | **Pass.** Zero syntax, zero schema change. |
| **P-11** intent vs execution | **Pass.** Names *who owes* the value, not *how* it is produced. |
| **P-12** 12-factor | **Pass.** Config stays in `env:`; this describes its origin. |
| **P-13** additive extensibility | **Pass.** No field changes meaning — with one honest caveat below. |

**Risks:**

1. **The precedence is a real decision, not a description.** For `required: true` + `generator:`, the spec previously said nothing and a tool could legitimately have prompted the operator. After this, that tool is non-conforming. No shipped file changes meaning, but a *tool's* behavior may have to. That is the one place "purely descriptive" would be too comfortable a claim.
2. **The classes are exhaustive by construction**, which means a future field (a `from:` pointing at an external secret store, say) has to be slotted into one of them or force a fifth. Worth knowing before ratifying four as complete.
3. **Reversibility: high.** Prose only; nothing to migrate.

### Open question for review

Should the classes be **surfaced by tooling** — a `validate` line naming each variable's class — or stay purely a spec concept? The argument for surfacing is that it makes the miniflux ambiguity visible to the author who wrote it. The argument against is that D-44 just added a `validate` audit line for host capabilities, and a second one starts turning `validate` into a report rather than a check.

ziadsawalha:
Names the four ways an env value can come to exist, and states what each layer owes for each. Records it as **D-49**.

Closes #181

## What it settles

Every declared env value falls into **exactly one** class, by an explicit precedence over fields that already exist — `generator:` → expression `default:` → literal `default:` → neither:

| Class | Trigger | Who owes the value |
|---|---|---|
| **Minted** | `generator:` present | The platform generates once, then preserves. A `generator:` **satisfies** `required:`. |
| **Derived** | `default:` is an expression over platform-resolved inputs | The provider, at resolution time |
| **Author default** | a literal `default:` | The file author chose a starting value |
| **User-supplied** | no generator, no default | Only a person can supply it |

The precedence is the load-bearing part. `miniflux`'s `ADMIN_PASSWORD` declares `required: true` **and** `generator: secret`; the spec said nothing about which wins, so a tool reading `required` first would prompt a human for a password the platform was about to generate. It classifies as **minted**.

## The defect this closes

**Two reference providers disagreed, in tree.** Docker and aws resolve `generator` → `default` → `required`. macos-dev filled the `default` first, and `resolveGenerators` skips any key already set — so `default` won there. Observed:

```
{ FOO: { default: "author-literal", generator: secret } }

macos-dev  FOO = author-literal
docker     FOO = 2b8109bb591bf75d6eda311c839ec480648c4a43bca5d1b6eb0af3bc7f9f0544
```

Same declaration, two values — a live P-5 violation. Fixed here, and **behaviour-neutral on everything shipped**: zero catalog or draft file declares both fields on one variable.

## Motivation

All four classes are already in the catalog — **93 declarations across all 72 apps** classify:

```
  minted            26 declarations / 23 apps
  derived           23 declarations / 20 apps
  author-default    34 declarations / 22 apps
  user-supplied     10 declarations /  5 apps
```

The classifier is published in the issue so these reproduce rather than being asserted.

## How the #181 verdict was addressed

Two `REJECT` rounds, six blockers, then `ACCEPT`:

| Blocker | Resolution |
|---|---|
| Counts didn't reproduce (86 vs 93) | Corrected, with the classifier published. The errors were skipping `components.*.env:` blocks and not treating a scalar shorthand as a `default:` — the review reproduced the original figures *per class* by reintroducing exactly those two omissions |
| "Spec and docs only, no code" was false | Scope corrected; the macos-dev fix ships with it, and the observed divergence now leads the motivation instead of the hypothetical |
| Two normative rules stated only in the draft | `default:` satisfies `required:`, and the `set_env` constancy rule, both pulled into the issue for review |
| `providers/macos-dev` failed `typecheck` | Fixed — Vitest was green because it doesn't typecheck, and CI runs typecheck first |
| "Minted once, then preserved" unmet by every provider | Recorded as present fact and tracked in **#186** |
| `#PROVENANCE` placeholder | → #181 |

## Known gap, recorded

The **preservation** half of the minted obligation is met only for the top-level `secrets:` block. `env:`-level generators are re-minted on every `up` — `resolveGenerators` takes no state, `LaunchState` has no env store, and Docker rebuilds its compose file each time. That is **26 declarations across 23 apps**, including this decision's own examples: `outline`'s `SECRET_KEY`, `miniflux`'s `ADMIN_PASSWORD`.

The obligation is stated because it is right; the gap is recorded because it is real. Tracked in **#186**. What is deferred is the work, not the obligation.

Also corrected on review: `generator: port` is **exempt** from preservation — a port is an allocation, not an identity, and preserving one across a host move produces a bind conflict.

## Tests

sdk 290/290 · macos-dev 108/108 · docker 103/103 · aws 40/40 · all typechecks clean

## Stacking

Based on `chore/gov-failure-semantics` (**D-48**).


ziadsawalha:
Both blockers fixed. `c87e6e7` (spec) and `9381c27` (www-dev).

## Blocker 1 — the user-supplied gap is now recorded — `c87e6e7`

I re-derived the divergence before writing the sentence, since it was going into permanent precedent. It reproduces, and it is slightly worse than described:

```
providers/docker/src/compose-generator.ts:733-742
  url|domain|origin  -> "http://localhost"
  email|mail         -> "test@localhost"
  else               -> "PLACEHOLDER"

providers/aws/src/translate.ts:967-973
  url|origin         -> "http://localhost"
  else               -> "PLACEHOLDER"

providers/macos-dev/src/env-writer.ts
  (no substitution — left unset)
```

The two guess tables **don't agree with each other**. Docker branches on `domain` and on `email`/`mail`; AWS does neither. So `ADMIN_EMAIL` on an unsupplied `required:` declaration is `test@localhost` on Docker, `PLACEHOLDER` on AWS, and unset on macos-dev — three values from one declaration, and `outline`'s `URL` (this decision's own class-4 example) splits two ways.

D-49's **Conformance at adoption** paragraph now records that alongside the minted-class gap it already carried, in the same form and the same place, tracked to #191/#192. The obligation stands; the gap is stated. Same posture you accepted for the minted class — I didn't invent a second one.

## Blocker 2 — the docs no longer teach a guarantee that isn't delivered — `9381c27`

Both lifecycle bullets in `www-dev/src/pages/learn/env-and-secrets.astro` are caveated rather than deleted, each pointing at the tracking issue:

- **Generated values** — *"Not yet true of inline `env:` generators in the reference providers: today a redeploy mints a fresh value. Tracked in #186."*
- **Values you supply** — *"Not yet true of the Docker and AWS providers, which today substitute a placeholder guessed from the variable's name for an unsupplied `required:` value. Tracked in #191 and #192."*

That satisfies #186's acceptance criterion — the docs line no longer runs ahead of the fix. It also caught the second bullet, which had the same defect for the user-supplied class and wasn't in your report.

## PR body counts — you were right, corrected

| Suite | Body claimed | Observed at `9381c27` |
|---|---|---|
| `sdk` | 290 | **290 pass, 0 fail** |
| `providers/macos-dev` | 105 | **108 pass, 0 fail** |
| `providers/docker` | 100 | **103 pass, 0 fail** |
| `providers/aws` | 40 | **40 pass, 0 fail** |

Body updated to 108 and 103. Run in build order, sdk built before providers.

## Not fixed here, and not claimed

- **CI has never run** — base is `chore/gov-failure-semantics`, `ci.yml` triggers on `pull_request: branches: [main]`, `statusCheckRollup` is `[]`. Clears when the train lands, this retargets to `main`, and a sync push follows. Nothing above claims CI coverage.
- **#198** — the 11 dead cross-section anchors. 10 of 11 predate this PR; agreed it is not a gate here.
- **#186 stays open.** This PR records and depends on it; it does not fix it. That framing is unchanged.

ziadsawalha:
Synced with the base in `fd30506`, since #193's branch had moved a long way — 23 commits, including its own sync with `main`.

No conflicts. Verified nothing was lost in either direction:

| Check | Result |
|---|---|
| Decisions present | D-44, **D-45**, D-46, D-47, D-48, D-49 |
| D-49 user-supplied gap sentence (this PR's blocker 1 fix) | present |
| `www-dev` env-and-secrets caveats (blocker 2 fix) | both present |
| `redact.ts` from #193 | arrived |
| `safeArgs` sites in `shell.ts` | 6 |

D-45 arrives here for the first time — its reservation stub landed on #176, which merged to `main`, and reaches this branch now via #193's sync.

Suites at `fd30506`, under CI's runner (`bun test`), build order sdk → providers:

| Suite | Result |
|---|---|
| `providers/docker` | **120 pass, 0 fail** |
| `providers/macos-dev` | **115 pass, 0 fail** |

All four typechecks clean. The counts moved from the ones in your APPROVE (290 / 103 / 108 / 40) because the merge brought `main` plus #193's redaction and wiring tests onto this branch — no test on this PR's own content changed.

Your APPROVE stands on content; this is a base sync, not new substance. Flagging the numbers explicitly so the delta isn't mistaken for drift.

Chain to `main` is now #194 → #193 → `main`, and **#193 is APPROVED, `CLEAN`, 12/12 green on `main`**. Once #193 lands this retargets and CI runs here for the first time.

PR #196, opened 2026-08-22
spec: a provider that does not execute schedule MUST say so at launch (D-51)

### LINKED ISSUE #190: RFC: a provider that does not execute `schedule` MUST say so at launch
## RFC: a provider that does not execute `schedule` MUST say so at launch

One conformance rule. **No new fields, no schema change, no syntax.**

### Problem

`schedule:` declares that a component runs on a timer. Nothing in the spec says what a provider owes when it has no scheduler — and the three reference providers each landed somewhere different:

| Provider | On a component with `schedule:` | |
|---|---|---|
| `@launchfile/docker` | `"<name>: has schedule — included but won't cron"` (`compose-generator.ts:507-509`) | warns |
| `@launchfile/aws` | a `nice-to-have` conformance gap, *"cron schedule not mapped"* (`translate.ts:939-947`) | reports |
| `@launchfile/macos-dev` | **nothing** — `grep -c schedule providers/macos-dev/src/provider.ts` → **0** | silent |

### Why silence is the wrong default here, specifically

Most unhonored fields announce themselves. Drop a port mapping and the app is unreachable; skip a volume and data vanishes on restart. An unexecuted `schedule` produces **no error, no missing endpoint, and no failed request** — the deploy succeeds, the component is up, and the only symptom is that a job silently never runs. There is nothing for the author to notice until they go looking for work that was never done.

So silence reads as success. That is the failure mode the existing conformance rule already names — *"report gaps, not silent drops"* (`PROVIDERS.md` §10 item 8) — applied to a field where the drop is invisible by construction.

### Proposed change

Extend §10 item 8 with the hard form for this field:

> A provider that does not execute a component's `schedule` **MUST** emit a launch-time warning naming the component.

Naming the component matters: in a multi-component app, "some schedule won't run" is not actionable.

The reference `macos-dev` provider gains the warning it lacked. Docker's existing warning already satisfies the rule; the AWS conformance gap already satisfies it for a translate-only provider.

### Motivation

Being straight about the numbers: **2 catalog entries declare `schedule:`** — `catalog/drafts/diun` and `catalog/drafts/nextcloud` — plus `spec/examples/cron-job.yaml`. Both are drafts; **0 of 72** shipped apps use it.

So this is not a widespread breakage. The case is:

1. **The one provider a developer runs locally is the silent one.** `macos-dev` is where an author first tries their Launchfile, and it is the only one of the three that says nothing.
2. **The spec ships an example teaching the field** (`spec/examples/cron-job.yaml`) with no provider that executes it. An author following the docs gets a deploy that reports success and does nothing on a timer.
3. **The rule is one sentence and the fix is one warning.** Low cost, and it closes a class where the cost of silence is unusually high.

If the steward considers 2 draft entries below the motivation bar, that is a fair call — though it would leave the spec teaching a field whose non-execution is undetectable.

### Principle self-assessment

| Principle | Assessment |
|---|---|
| **P-5** provider-translatable | **The point.** Three providers, three behaviors, for a field whose non-execution the author cannot otherwise detect. |
| **P-6** it's just YAML | **Pass.** Zero schema, zero syntax. |
| **P-11** intent vs execution | **Pass.** The rule binds *reporting*, not mechanism. A provider may still decline to schedule; it must not do so quietly. |
| **P-13** additive extensibility | **Pass.** No field changes meaning; every existing file stays valid. |
| **D-40** | Consistent — a diagnostic surface, not a hard failure. |

**Risks:**

1. **A warning is not execution.** This makes the gap visible; it does not make `schedule` work anywhere. An author who wanted a cron job still has to solve it outside the Launchfile. That is honest, but it is a smaller win than "schedule is supported".
2. **Warning fatigue** if a provider that legitimately never schedules is used with a scheduled component repeatedly. Mitigated by it being launch-time and component-named rather than per-invocation noise.
3. **Reversibility: high.** One conformance sentence and one warning.

### Open question for review

Should this generalize now rather than later? The same argument — *"the drop is invisible, so silence reads as success"* — applies to any field whose non-execution produces no observable symptom. Stating it as a general rule ("a provider MUST warn when it does not honor a field whose omission is not otherwise observable") would be broader but harder to conform to, since "not otherwise observable" is a judgement. Scoping to `schedule` is the narrow, checkable version; I lean narrow, but the general rule is the one that would prevent the next instance.

ziadsawalha:
A provider that does not execute a component's `schedule` **MUST** say so at launch, naming the component. Records it as **D-51**.

Closes #190

## What it settles

`PROVIDERS.md` §10 item 8 — *"report gaps, not silent drops"* — is already a MUST. This names `schedule` as its hard case, requires the warning name the component, and distinguishes known-but-unexecuted from the P-6 unknown-key tolerance.

**The `macos-dev` warning is a conformance bug fix, not a new requirement** — that provider violates item 8 on `main` today (`grep -c schedule providers/macos-dev/src/provider.ts` → **0**).

## Why silence is the wrong default for this field specifically

Most unhonored fields announce themselves — drop a port and the app is unreachable. An unexecuted `schedule` produces no error and no missing endpoint. The component is started once at launch, which reads as a successful first run; nothing afterwards distinguishes a scheduled component from an unscheduled one.

On `macos-dev` it is worse: `launchStatus()` prints components from the state file with **no process-liveness check**, so a one-shot job that ran and exited renders identically to a healthy long-running server.

Three providers, three behaviors on `main`: docker warns, aws emits a `nice-to-have` gap, macos-dev says nothing.

## Motivation — corrected after review

**The catalog evidence in the original RFC does not hold, and I want that on the record rather than buried.** I cited `catalog/drafts/diun` and `catalog/drafts/nextcloud` as motivating entries. The review checked both against upstream:

- **diun** declares `schedule:` *and* `DIUN_WATCH_SCHEDULE` *and* `restart: always`. Upstream: *"The `schedule` setting makes Diun run as a daemon."* **It schedules itself.**
- **nextcloud `cron`** runs `cron.sh`, which upstream is `exec busybox crond -f`. **A foreground cron daemon inside the container.**

So **0 of the 2 cited entries depend on provider-side scheduling**, and `0 of 72` shipped apps declare the field at all.

The motivation is therefore the spec's own `spec/examples/cron-job.yaml` — a registered, live docs example teaching a field with no executor — plus the three-way provider divergence. That is a spec-integrity argument, not a catalog-breakage one.

**The `3+ real-app` bar does not apply**: it gates *format additions*, and this adds no format surface. The applicable bar is the one item 8 already sets, which `macos-dev` fails today.

## How the #190 verdict was addressed

`ACCEPT` for the rule, with three blockers for this PR:

| Blocker | Resolution |
|---|---|
| **The warning text was factually false** — it claimed "nothing will run it on a timer", which is wrong for both self-scheduling entries | Both providers' warnings now state what *this provider* does and leave the app's own behaviour to the app. A conformance warning that misstates the user's app is worse than the silence it replaces |
| **D-51 cited D-40 for its surface** — but D-40 fences its diagnostic to `validate` *only, never operational commands* | Now cites rule 9 / D-44, the operational-surfacing precedent, and states why the two land differently (D-40 fires across the whole image-first catalog; this fires on under 1% of apps) |
| **No test fenced the behavior** — grep across all three providers' test dirs returned nothing | Seven tests across both providers, including one asserting the text does *not* claim the job will not run |

## Also on the record

The site over-claimed before this branch: `examples/[slug].astro` said *"the scheduler handles re-invocation"* and `learn/specialized-patterns.astro` said the task *"is started"* on schedule. Both asserted execution no provider delivers. Rewritten.

## Tests

docker 107/107 (+4) · macos-dev 111/111 (+3) · sdk 290/290 · aws 40/40 · typechecks clean

## Stacking

Based on `chore/gov-provenance` (**D-49**). `DESIGN.md` has a **D-50 hole** — that number is reserved for a decision being rebuilt after a `REJECT`. If D-50 does not land with this, the published governance page shows a gap.


ziadsawalha:
Blocker 1 fixed in `da06123`. The constraint is now in the contract, not only in the implementations.

## What rule 8 says now

The old sentence required only *"a launch-time warning naming the component"*. A third-party provider satisfied that by emitting `nothing will run it on a timer` — false about both catalog entries that declare the field, and the exact string both reference providers were scoped away from in this PR. The constraint lived in two source files and seven tests; the contract left it unbound.

Rule 8 now carries it:

> …a provider that does not execute a component's `schedule` MUST emit a launch-time warning naming the component **and the field** (D-51); a translation-only provider records the same gap in its conformance report, which is its equivalent of launch. **The warning states what *this provider* does and MUST NOT assert what the app will do** — a component may schedule itself, and a warning that misstates the author's app is worse than the silence it replaces.

Three things changed, deliberately:

1. **The content constraint is now normative**, which is what #190's blocker 1 asked for.
2. **The absolute phrasing is gone.** "nothing the author could notice at launch" was false — the component *is* started once at launch, and that is observable. What is missing is any signal about *future* runs. The replacement says that, and matches the form D-51's **Why** in `DESIGN.md` already carries, so the two documents no longer disagree.
3. **Translation-only providers are named.** The rule was phrased around "launch", which the AWS probe never performs — it records the gap at translate time. Rule 8 now says that satisfies it, rather than leaving a conforming provider unable to tell.

Also dropped a stray double blank line before D-42 that had pulled an unrelated region of `DESIGN.md` into this branch's diff.

## Verification — OBSERVED

Run at `da06123` in the branch worktree, in build order (sdk → providers):

| Suite | Result |
|---|---|
| `providers/docker` | **107 pass, 0 fail** |
| `providers/macos-dev` | **111 pass, 0 fail** |

**PR body corrected.** You were right that `docker 103/103 (+4)` was the pre-PR count — 103 is what `main` has; this branch is at **107**. The body now reads 107 and picks up the missing aws line.

## Blocker 2 — still stands, and I am not claiming otherwise

No CI has run on this content and none will before the train moves:

```
$ gh pr checks 196
no checks reported on the 'spec/schedule-honesty' branch
```

Base is `chore/gov-provenance`; `ci.yml` triggers on `pull_request: branches: [main]`. This clears when the upstream train lands, this retargets to `main`, and a sync push follows — the retarget alone fires none of `opened` / `synchronize` / `reopened`.

Worth flagging for when you re-review: `DESIGN.md` on this branch reads D-49 → **D-51**, and the missing **D-50** originates here. D-45 got a reservation stub on #176 (`62bc959`) on exactly this reasoning — the hole is the introducing PR's artifact and becomes permanent public record on merge. Same treatment wanted here, and it is cheaper now than after. Say the word and it goes in this branch.

Nothing in this push touches #199 or #200 — both remain the right homes for the execution gap and the draft defects.

ziadsawalha:
Correcting one line in my previous comment: I wrote that "103 is what `main` has". It isn't — I checked afterwards rather than before, and got it wrong.

```
origin/main                        ->  89 pass, 0 fail
origin/chore/gov-provenance        -> 103 pass, 0 fail   <- this PR's base
origin/spec/schedule-honesty       -> 107 pass, 0 fail   <- this PR
```

103 is the count at this PR's **base**, not at `main`. The `+4` in the body was right all along; only the total was stale, and the corrected `107/107 (+4)` stands. The 89 → 103 difference is the four upstream PRs in the stack, not anything on this branch.

Nothing else in that comment changes — the rule 8 rewrite, the four suite results, and the CI status were all run before posting.

ziadsawalha:
D-50 stub added in `bfab49d`, taking your framing rather than D-45's.

## The stub

```markdown
### D-50: Reserved — no decision recorded

**Status**: Reserved; no decision recorded. No proposal claims this number — it was
vacated while this branch was open, when the decision now recorded as D-51 was
renumbered twice (D-43 → D-50 → D-51) …
```

Your distinction is the right one and I verified the history before writing it:

```
$ git log --oneline --all | grep -E '^(7ac26a0|f428ec6)'
f428ec6 spec: renumber the schedule-warning decision to D-51
7ac26a0 spec: renumber the schedule-warning decision to D-50
```

So D-50 was reserved for **nothing** — unlike D-45, which is genuinely held for #171. Saying "held for #N" here would have invented a reservation that does not exist, which is a worse defect than the hole. The stub says vacant, and says why.

Agreed on not renumbering: it would be D-51's third renumber and #197 already cites both numbers.

`grep -oE '^### D-(4[4-9]|5[0-2])'` on this branch now returns D-44, D-46, D-47, D-48, D-49, **D-50**, D-51 — contiguous from 46. D-45 is absent here only because its stub landed on #176 and reaches this branch when the train syncs from `main`; D-52 is #197's.

## Also landed — `cd90b92`

Merged `chore/gov-provenance` in, since #194's D-49 fixes had moved the base. Clean, no conflicts. Verified nothing was lost in either direction:

| Check | Result |
|---|---|
| D-49 user-supplied gap sentence | present |
| D-50 stub | present |
| rule 8 `MUST NOT assert what the app will do` | present |
| `www-dev` env-and-secrets caveats | both present |
| `sdk` / `docker` / `macos-dev` / `aws` (`bun test`) | **290 / 107 / 111 / 40**, 0 fail |

## One thing to ignore

**GitHub reports this PR as `DIRTY`, and it is wrong.** The head is a strict descendant of the base, so the merge is a fast-forward and cannot conflict. Three independent checks:

```
$ git merge-base --is-ancestor origin/chore/gov-provenance origin/spec/schedule-honesty; echo $?
0
$ git rev-list --count origin/spec/schedule-honesty..origin/chore/gov-provenance
0
$ git merge origin/spec/schedule-honesty --no-commit --no-ff   # from a checkout of the base
Automatic merge went well; stopped before committing as requested
```

Zero conflicted paths. A REST re-query of `pulls/196` did not shake the cache loose. Flagging it so it is not mistaken for a real conflict when the train moves.

## Still open

Blocker 2 (CI) is unchanged and unclaimed — no `ci.yml` job has run here. It clears as the train lands: #176 is merged, #177 is green and APPROVED, and this PR's base chain has to reach `main` before the filter matches.

ziadsawalha:
Suggestion applied in `c7c1693`. Both points were right.

`spec/DESIGN.md:563` now reads:

> …it was vacated while [#196](https://github.com/launchfile/launchfile/pull/196) was open, when the decision now recorded as D-51 was renumbered twice (D-43 → D-50 → D-51) **to clear numbers already claimed by other decisions** and to match the number [#197](https://github.com/launchfile/launchfile/pull/197) carries for it.

Naming #196 is the fix — "this branch" is meaningless to a reader of launchfile.org, and D-45's stub already set the pattern by naming #171.

## One correction to your correction

Your reasoning for dropping the `main` attribution is right, but the replacement attribution isn't quite. I checked before committing, and the branch that held D-43 was not the host-capability one:

```
$ git show -s --format=%ci 7ac26a0
2026-08-22 21:00:54 +0200

$ MAINAT=$(git rev-list -n1 --before="2026-08-22 21:00:54 +0200" origin/main)
$ git log --oneline -1 $MAINAT
3d464d5 chore: upgrade Changesets to CLI v3 for changesets/action v2
$ git show $MAINAT:spec/DESIGN.md | grep -c '^### D-43'
0

$ git log --oneline --all -S'### D-43:' -- spec/DESIGN.md | tail -1
772ab4f spec: define source acquisition — repository as canonical origin, #ref baseline fragment (D-43)
```

So: `main` genuinely had **no** D-43 at that moment — your central point stands, and "to clear numbers taken on `main`" was wrong for the first renumber exactly as you said. But the number was held by the **source-acquisition** decision (`772ab4f`), not by a host-capability branch allocating 43–48 as a block.

This is why the generic wording is the right landing place. "Numbers already claimed by other decisions" is true of both renumbers, true at every point in time, and does not require the reader to trust a branch-archaeology claim that neither of us had quite right. I'd rather the published sentence assert less and be durable than assert more and need a future correction.

My commit message for `c7c1693` repeats your host-capability framing — that part is unverified and I'm flagging it here rather than rewriting the commit. The `DESIGN.md` text itself carries no such claim.

## Everything else

Nothing outstanding on content. Agreed the remaining gate is CI, and that it is structural: this base is `chore/gov-provenance`, and the chain to `main` is #196 → #194 → #193 → `main`. **#177 has since merged and #193 is now APPROVED, `CLEAN`, 12/12 green on `main`** — so the chain is two merges from reaching this PR.

Your `merge-tree --write-tree` check on `cd90b92` is a better proof than the one I offered — showing git's own auto-merge tree is byte-identical to what was committed rules out hand edits entirely, which my "no conflicts" check did not. Same for the DIRTY proof: pointing out that GitHub's reported base SHA *is* the second parent of its reported head SHA is self-refuting in a way my three commands weren't.

ziadsawalha:
**Retracting my "ignore the DIRTY flag" advice.** It was correct when I wrote it and stopped being correct shortly after, which is worse than being wrong once — anyone acting on it later would have merged past a real conflict.

What changed: I synced `chore/gov-provenance` with #193's branch in `fd30506`. That moved this PR's base 24 commits ahead, so the head stopped being a descendant and the fast-forward argument evaporated. My original three commands were sound about the tree at that moment; none of them stay true as the base moves. That is the actual lesson — a mergeability proof has a shelf life measured in pushes, and I presented one as a standing fact.

Resolved in `29278e0`. `mergeable` is now `true`.

## The conflict

`providers/macos-dev/src/provider.ts`, exactly as you diagnosed: two unrelated new exported functions at the same anchor after `refusedHostCapabilities` —

- `scheduleWarning` (D-51, this branch)
- `applyHostCapabilityRefusals` (D-44, relocated on the base)

Neither supersedes the other. Kept both.

| Check | Result |
|---|---|
| `providers/macos-dev` `bun test` | **118 pass, 0 fail** |
| `providers/macos-dev` `bun run test` | **118 passed** |
| `providers/docker` | **124 pass, 0 fail** |
| typechecks (all four) | clean |

118 is this branch's 111 plus the 7 refusal-fencing tests from the base — matches your independent resolution exactly.

`DESIGN.md` now runs **D-40 → D-51 with no holes**: D-45's stub arrives from `main` via the base, D-50's is this branch's own. Both reservations are doing their job.

## On D-43 — your correction is better than my derivation

The duplicate `### D-43` in `886c4e6`'s tree settles it without timestamps or branch archaeology. Mine relied on `rev-list --before` against a commit whose committer date is a rebase artifact, and you are right that it no longer reproduces as `main` grows — which makes it exactly the kind of evidence that should not have been offered as proof.

That strengthens rather than weakens the case for the generic wording now in `DESIGN.md`. A sentence whose accuracy depends on which date field a future reader picks does not belong in permanent public precedent.

## Remaining

Your step 4 is the one that matters and it is now done: this head carries the resolved sync. What is left is upstream — merge #193, then #194 retargets and merges, then this retargets to `main` and needs one more push to fire `synchronize`, since a base change alone fires none of the default trigger types.

#197 is stacked on this head and will need the same carry-forward.

PR #197, opened 2026-08-22
spec: a provider MUST NOT fabricate a value for an unsupplied required env var (D-52)

### LINKED ISSUE #191: RFC: a provider MUST NOT fabricate a value for an unsupplied required env var

ziadsawalha:
A provider **MUST NOT** substitute a fabricated value for an unsupplied `required` env var. Records it as **D-52**.

Closes #191

## The defect

`env.<NAME>: { required: true }` is the app's own statement that it cannot start without the value. The reference Docker provider invents one, guessing from the variable's **name** (`compose-generator.ts:728-737`):

```ts
if (lowerKey.includes("url") || lowerKey.includes("domain") || lowerKey.includes("origin"))
  return "http://localhost";
if (lowerKey.includes("email") || lowerKey.includes("mail"))
  return "test@localhost";
return "PLACEHOLDER";
```

Run against the shipped catalog:

```
outline:  URL: http://localhost              warnings: []
posthog:  SITE_URL: http://localhost         warnings: []
flowise:  FLOWISE_USERNAME: PLACEHOLDER
          FLOWISE_PASSWORD: PLACEHOLDER      warnings: []
```

`warnings: []` is the load-bearing part — **the substitution is undiagnosed**. `flowise`'s `FLOWISE_PASSWORD` is `sensitive: true`, so every deployment gets the same publicly-known admin password (D-18).

The guess is wrong *in kind*, not just in value. In drafts: `supabase` gets `PGRST_DB_URI=PLACEHOLDER` and `GOTRUE_DB_DATABASE_URL=http://localhost` — an HTTP URL in a Postgres DSN slot. `fider` gets `EMAIL_SMTP_HOST=test@localhost` — an email address in a hostname slot.

**This is worse than failing.** A failure names the missing variable. `http://localhost` as a public URL is a silent misconfiguration that surfaces later as behaviour nobody connects back.

## The rule

Keyed to the verb, because "fail" is not available everywhere:

| Verb | Obligation |
|---|---|
| **Deploying** (`up`, anything that provisions or runs) | Obtain from the provider's operator channel, or **fail naming the variable**. A **prompt** counts only in an interactive session — a non-interactive invocation MUST fail, or a fast named failure becomes a CI job hanging on stdin |
| **`translate`** | Leave the variable **absent** from the artifact and record it unmet |
| **Inspection** (`env`, `status`, `list`, `logs`) | Report it unsupplied rather than omitting it or showing an invented value |

## How the #191 verdict was addressed

`REJECT` with three blockers, all mechanical:

| Blocker | Resolution |
|---|---|
| **Two content-free `### D-` headings in `DESIGN.md`** — an empty `D-43` colliding with the real one, and an empty `D-50`. Rebase debris. `www-org` renders `DESIGN.md` whole, so **both would have published to launchfile.org** | Removed. Verified: 50 decisions, no duplicates, none empty |
| Duplicated stale `D-50` comment in `providers/macos-dev/src/provider.ts` | Removed |
| **The conformance gap was stated but not tracked** — D-49 links #186 for the same posture; this linked nothing | Now **#192** |

The review also corrected my counts: 3 shipped apps was right, but **12 of 41 drafts and 2 `spec/examples/` files** are affected too — including `single-component.yaml`, the spec's canonical teaching example.

## Known non-conformance, recorded and tracked

**When this lands, no deploying provider implements the fail branch.** `@launchfile/aws` implements the `translate` branch here. Docker still fabricates; macos-dev fails **both** the deploying and the `env` inspection branches.

Bringing Docker into conformance needs a catalog pass first — 3 apps, 12 drafts and 2 spec examples currently depend on the fabricated values to launch.

One hazard worth naming precisely: `catalog/test/src/launch-to-compose.ts` carries an **independent copy** of the same heuristic. So nothing breaks when this lands — and that is the problem. Once Docker conforms, the harness will still fabricate, and `health_check_passed: true` will certify apps the shipped provider refuses to deploy. `outline`'s metadata already records that stamp for an app configured with an invented public URL.

All of it is **#192**. What is deferred is the work, not the obligation.

## Composes with D-49

D-49 (in this same stack) classifies `outline`'s `URL` as **user-supplied** and states a `generator:` or supplied `default:` satisfies `required:` — which is exactly this rule's arrival test. D-52 is the provider-side consequence of a declaration-layer rule D-49 states. No conflict.

## Tests

sdk 290/290 · docker 107/107 · macos-dev 111/111 · aws 50/50 · all typechecks clean · AWS conformance ledger regenerated and `--check` passing

The 10 AWS tests cover the arrival-test edges — a binding on an unmappable resource, a `supports`-only binding, and reporting when the component gaps for an unrelated reason. Whoever brings Docker into conformance should port that table.

## Stacking

Top of the train, based on `spec/schedule-honesty` (**D-51**).


ziadsawalha:
Blocker 2 fixed in `f24e546`. Took the restructure, not the count bump.

## Why (b), not (a)

You were right that the prompt bullet is a qualifier rather than a branch, and that settles which fix is honest. Changing "four" to "five" would have made the prose agree with a list that mis-classifies its own contents — the partition would still read as five peer branches when it is four, and a future reader checking §2's verbs against it would find only four things to check against.

So the prompt sentence is now folded into the deploying branch it qualifies, and the list is genuinely four:

```
- **Deploying verbs** (`up`, and any verb that provisions or runs) …
- **`translate`**, where there is no launch at which to fail …
- **Verbs that report a component's resolved environment** (`env`; `status` where …)
- **`down`, `list`, `logs`, and a `status` whose output does not include resolved environment**
```

The prompt rule loses nothing — it stays normative, sits directly against the branch it constrains, and reads better there, since "a prompt counts as an operator channel only in an interactive session" only ever made sense as a constraint on *that* branch's operator-channel clause.

## Blocker 1 — CI — stands

```
$ gh pr checks 197
no checks reported on the 'spec/required-env-honesty' branch
```

Tip of a six-PR stack, so it clears last: train lands, retarget to `main`, sync push.

**Narrowing evidence on `aws-terraform-validate`, explicitly not discharging the gate.** `validate-tf.ts:35` honors `TERRAFORM_BIN`, and OpenTofu is installed here:

```
$ TERRAFORM_BIN=tofu bun run validate:tf --examples
…
13/13 valid (tofu)
```

Same script, same emitted HCL, different binary. Strong signal that the Terraform this PR affects parses and schema-validates — but it is not the recorded gate, and I am not treating it as one.

## Verification — OBSERVED, under CI's runner (`bun test`), build order sdk → providers → CLI

| Suite | Result |
|---|---|
| `providers/docker` | **107 pass, 0 fail** |
| `providers/macos-dev` | **111 pass, 0 fail** |
| `providers/aws` | **50 pass, 0 fail** |
| `packages/launchfile` | **42 pass, 0 fail** |
| `bun run src/conformance.ts --check` | `CONFORMANCE.md is up to date` |

Worth carrying over from #193: **`bun test` and `bun run test` are different runners in this repo**, and `ci.yml` uses `bun test` while the package `test` script is `vitest run`. A new test file on #193 passed under vitest and failed under bun; the numbers above are from CI's runner.

## #192 extended, not duplicated

Verified your finding and added the detail to #192 rather than filing anything new. `librechat`'s `OPENAI_API_KEY` and `ANTHROPIC_API_KEY` and `supabase`'s `DB_PASSWORD` are all `required: true` + `sensitive: true`, and none matches the `url`/`origin`/`domain`/`email` branches in either fabrication table — so all three fall through to bare `PLACEHOLDER`, same as `flowise`. Argued there that fabricated **credentials** belong in the first conformance pass rather than a later one, since a wrong key fails at an auth boundary rather than fast and visibly.

## D-50 — agreed, and it belongs on #196

Confirmed absent at this head:

```
$ git show origin/spec/required-env-honesty:spec/DESIGN.md | grep -cE '^#+ D-50'
0
```

D-45 got a reservation stub on #176 (`62bc959`) on exactly this reasoning: the hole is the introducing PR's artifact, and it becomes permanent public record on merge. **D-50 arrives from #196**, so the stub wants to land there before #196 merges — not here, and not as a follow-up afterwards. Raised it on #196.

Head is `f24e546`.

ziadsawalha:
Heads-up: this PR went `DIRTY` after the #196 fix landed on its base, and is now merged clean again in `b719767`.

Both branches rewrote `PROVIDERS.md` rule 8 — this one restructured it into a two-field list (`schedule` + required env, D-51 + D-52), while #196 rewrote the single-field form to bind the D-51 warning-content constraint. Git could not reconcile them.

Resolved to **this branch's two-field structure carrying #196's corrected schedule wording**. That combination matters: this branch's `schedule` sub-bullet still read *"MUST emit a launch-time warning naming the component"* with no content constraint, and still carried the absolute *"nothing the author could notice at launch"* that `DESIGN.md` was corrected away from. Taking either side wholesale would have silently dropped one of the two fixes — #196's constraint, or D-52 entirely.

Verified after the merge:

```
$ grep -c '^<<<<<<<' spec/PROVIDERS.md spec/DESIGN.md
0
0
$ grep -cE '^#+ D-5[12]' spec/DESIGN.md
2
$ grep -nE '^[0-9]+\. \*\*' spec/PROVIDERS.md | tail -4
215:8. **Report gaps, not silent drops** …
231:9. **Grant or refuse `host:` capability entries** …
232:10. **Honor the failure semantics** …
233:11. **Interpret command strings with a POSIX shell** …
```

Rule numbering unbroken, both decisions present, no markers left. `spec/DESIGN.md` auto-merged with no conflict.

Suites at `b719767`, under CI's runner (`bun test`), build order sdk → providers:

| Suite | Result |
|---|---|
| `providers/docker` | **107 pass, 0 fail** |
| `providers/macos-dev` | **111 pass, 0 fail** |
| `providers/aws` | **50 pass, 0 fail** |

`mergeStateStatus` is back to `CLEAN`.

This is worth watching as the train moves: #196 and #197 both edit rule 8, so any further change to it upstream will conflict here again. The `schedule` sub-bullet is the shared line.

ziadsawalha:
Blocker fixed in `4d24bed`, and the base is synced in `1139929`.

## The caveat this PR was about to falsify

You were right, and the reasoning you used is the one I argued for the D-50 stub — a statement the merge makes false is the landing PR's artifact, not a follow-up. Verified at this head:

```
$ grep -rn "PLACEHOLDER\|http://localhost\|test@localhost" providers/aws/src/*.ts
(no matches)

$ grep -n "PLACEHOLDER" providers/docker/src/compose-generator.ts
741:		return "PLACEHOLDER";
```

AWS fabrication is gone; Docker's remains. `www-dev/src/pages/learn/env-and-secrets.astro:107` now reads:

> *Not yet true of the **Docker provider**, which today substitutes a placeholder guessed from the variable's name for an unsupplied `required:` value. **The AWS provider no longer does.** Tracked in #191 and #192.*

I kept the explicit "no longer does" rather than silently dropping AWS. The page's job here is to tell a reader what the providers actually do, and "AWS was fixed" is the more useful sentence than an unexplained absence — particularly since #192 still lists AWS in its scope.

## Base sync — `1139929`

25 commits from #196, no conflicts. **D-45 has arrived** — it reaches here via #196's own base sync (`29278e0`), which is one step earlier than "when the train reaches `main`", exactly as you predicted. `DESIGN.md` now carries D-45 **and** D-50, so both reservation stubs are in place on this head.

## Verification — OBSERVED, CI's runner

| Suite | Result |
|---|---|
| `providers/docker` | **124 pass, 0 fail** |
| `providers/macos-dev` | **118 pass, 0 fail** |
| `providers/aws` | **50 pass, 0 fail** |
| typechecks (all four) | clean |
| `www-dev` `astro check` | **29 files, 0 errors, 0 warnings, 0 hints** |

Counts moved from the numbers in your review (290/107/111/50) because the sync brought #193's redaction work and #196's refusal-fencing tests onto this head. No test on this PR's own content changed.

One thing worth recording so nobody else loses time on it: `www-dev`'s typecheck first failed here with `Cannot find module 'satteri'`. That is not a defect and not in any diff — `satteri` is declared in `www-shared/package.json`, and that workspace simply had no `node_modules` in this worktree. `bun install` in `www-shared/` fixes it. The `astro check` result above is after that.

## Remaining

CI, unchanged, and your two routes are both still open. Worth noting the chain got shorter while you were reviewing: #196's conflict is resolved (`29278e0`, `mergeable: true`), so it is now #197 → #196 → #194 → #193 → `main`, with #193 and #194 both APPROVED and green.

On your run-`32606489278` anomaly — agreed it should not be read as evidence that stacked PRs get CI. #194's matrix ran because its base was `chore/gov-failure-semantics` at a moment when that branch had `main` merged into it, not because the filter matched a stacked base. The `pull_requests` array resolving live is a good caution to have on the record.

ziadsawalha:
**Two corrections, and the CI blocker is discharged.**

## 1. My explanation of your run-`32606489278` anomaly was wrong

I wrote that #194's matrix ran "because its base was `chore/gov-failure-semantics` at a moment when that branch had `main` merged into it." That is not an explanation — the `branches:` filter matches the base **ref name**, not its content — and you had already said the base at trigger time is unrecoverable. I asserted a mechanism you had explicitly declined to conclude, which is exactly the thing not to do.

My second theory was also wrong. I thought #195's merge explained it. The timing refutes that:

```
$ gh pr view 195 --json mergedAt                              -> 2026-08-23T00:01:28Z
$ gh api .../actions/runs/32606489278 --jq .created_at        -> 2026-08-22T23:54:56Z
```

Six minutes *before*. At 23:54 every `ci.yml` in reach still carried `branches: [main]`, and #194's base was `chore/gov-failure-semantics`. **I do not know why that run fired.** Your caution stands and my two attempts to resolve it both failed — recording that rather than offering a third.

## 2. The CI blocker is now genuinely discharged — but not for the reason I gave

**#195 merged at 00:01:28**, deleting the base filter. `main`'s `ci.yml` now opens *"Every pull request, whatever its base."* The first `ci.yml` run on each stacked branch lands after that merge, which is the part that *is* explained:

```
chore/gov-provenance        2026-08-22T23:54:56Z   <- the unexplained one
spec/schedule-honesty       2026-08-23T00:08:56Z   <- after #195
spec/required-env-honesty   2026-08-23T00:14:07Z   <- after #195
```

Current state of the train — all four now carry the full 13-check matrix:

| PR | Decision | Merge | Checks |
|---|---|---|---|
| #193 | APPROVED | CLEAN | 13 |
| #194 | APPROVED | CLEAN | 13 |
| #196 | CHANGES_REQUESTED | CLEAN | 13 |
| #197 | CHANGES_REQUESTED | UNSTABLE | 13 |

So the single item your last two reviews held on — *"`ci.yml` has physically never run against this tree"* — no longer holds. `aws-terraform-validate`, the casualty you named specifically, is in that matrix.

This PR is `UNSTABLE` only because one check is still running at the time of writing; the rest pass.

Worth stating plainly since it bears on your earlier reasoning: **the retarget-and-push route was never needed for #196 or #197.** #195 was the fix, and landing it propagated to every stacked PR without any of them being retargeted. Your round-2 review on #177 correctly said #195 was orthogonal *for that PR* — which was true, because #177 had already retargeted to `main`. It was not orthogonal for the ones still stacked.

PR #249, opened 2026-08-24
Speed up the AWS terraform-validate CI job (93s → seconds)

ziadsawalha:
`AWS provider — terraform validate` is the longest job in CI: 101s in [run 32711412319](https://github.com/launchfile/launchfile/actions/runs/32711412319), 93s of it in one step.

**Cause.** `validate:tf` ran `terraform init` once per example directory. Measured with OpenTofu on a warm plugin cache: init ~17s, validate ~1.2s. So init was ~93% of the per-example cost, paid 13 times, sequentially. The job log shows this as a flat ~7.2s per example — flat is the tell, since a download would spike only the first.

**Change.** Group sources by their emitted `terraform { required_providers { … } }` block, run one `init` per group, symlink `.terraform` into the rest, and run the validates concurrently (`VALIDATE_TF_CONCURRENCY` to override). Also drop `needs: sdk` from the job — it builds the SDK itself in ~1s, so the dependency only added 15s of queue latency to the run's longest pole.

**Measured locally** (OpenTofu, warm cache): `--examples` 229.6s → 21.7s; full catalog (85 sources) 32.4s.

**The gate still gates.** With a `bad.tf` carrying an unsupported argument injected into every directory, the run reports `0/13 valid`, exits 1, and each failure cites the AWS provider schema. A borrowed `.terraform` resolves the real schema exactly as a per-directory init does.

The CI numbers in this PR are a projection from local timings — this PR's own run is the measurement.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,6 +27,17 @@ updates:
       interval: weekly
       day: monday
     open-pull-requests-limit: 5
+    ignore:
+      # TypeScript 7 (the native port) exports no `ts.sys`, and
+      # @astrojs/language-server calls `ts.sys.fileExists` when it looks for a
+      # tsconfig — so `astro check` dies with "Cannot read properties of
+      # undefined (reading 'fileExists')". @astrojs/check 0.9.10 is the newest
+      # release and still pins language-server ^2.16.7, so there is no version
+      # to upgrade into. These two sites stay on TypeScript 6 until a
+      # language-server release supports the 7.x API; the workspace root is
+      # unaffected because it runs `tsc` directly. Drop this once that lands.
+      - dependency-name: typescript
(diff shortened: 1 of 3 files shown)


ziadsawalha:
Measured on this PR's own run ([32715606739](https://github.com/launchfile/launchfile/actions/runs/32715606739)), against the baseline run [32711412319](https://github.com/launchfile/launchfile/actions/runs/32711412319):

| | before | after |
|---|---|---|
| `Validate emitted Terraform` step | 93s | **18s** |
| `AWS provider — terraform validate` job | 101s | **25s** |
| whole CI run, wall clock | 120s | **77s** |

Still `13/13 valid (terraform)`. The job is no longer the longest pole — `Website (www-dev)` at 62s now sets the run's wall clock.

PR #313, opened 2026-08-27
catalog+spec: listmonk release command, rallly and glance health checks

ziadsawalha:
Two commits.

## `catalog:` — listmonk did not deploy; two apps had no health check

**listmonk was broken three ways.** All observed with the reference Docker provider (Docker Desktop 29.7.2, `listmonk/listmonk:latest` v6.2.0, linux/arm64) — not inferred from source:

| Defect | Evidence | Fix |
|---|---|---|
| No `release`, so the schema is never installed | `main.go:159: the database does not appear to be setup. Run --install.` repeating in a restart loop until the 120s health gate expires — this is `main` today | `commands.release` running listmonk's own documented install/upgrade |
| `--config ''` reads config from env only; listmonk sets no `db.ssl_mode` default and omits empty fields from the DSN (`cmd/init.go`), so `sslmode` never reaches the connection string | release exits 1 on `pq: SSL is not enabled on the server` | `LISTMONK_db__ssl_mode: disable`, which upstream's own compose pairs with the same command |
| `health: /api/health` sits behind the auth middleware (`cmd/handlers.go`) | `403 {"message":"invalid session"}` | `health: /health`, upstream's "Public health API endpoint" |

Verified end to end: release completes, the health gate passes, `GET /health` returns `200 {"data":true}`, `GET /admin/login` returns `200`. A redeploy over the installed database logs `skipping install as database appears to be already setup` and `no upgrades to run`, and exits 0 — which is what `--idempotent` buys, and the claim the spec paragraph makes.

`metadata.yaml` is a fresh harness record. It reads `health_check_passed: false` because the catalog harness drops `commands.release` (no reference to it anywhere in `catalog/test/src`), so it reproduces the very crash loop this PR fixes. Only `notes` is hand-edited, to say so.

**rallly** — `health: /api/status`. Upstream defines an unauthenticated `GET /api/status` (`apps/web/src/app/api/status/route.ts`); its own compose health-checks only the DB. Verified live: `200`, `"database":"connected"`.

**glance** — `health: /api/healthz`. Upstream registers it (`internal/glance/glance.go:458`). **Not verified live:** the entry mounts its volume at `/app/data` while the image entrypoint is `--config /app/config/glance.yml`, so glance crash-loops before it serves. That is a separate pre-existing defect, not touched here.

## `spec:` — make the release-stage init pattern legible

One paragraph in SPEC's Commands section. One-time initialization that must complete before the app serves belongs in `release`, guarded by the app's own idempotent flags so it is safe to run on every deploy.

It states `release`'s own failure semantics rather than borrowing `bootstrap`'s. Both stages carry the same idempotency recommendation, but D-48 separated them on disposition: `bootstrap` is user-invoked and non-deploy-failing, while `release` runs automatically and a non-zero exit fails the deploy.

listmonk is the worked example, and the first `catalog/apps/` entry to declare `release`.

Kept as two commits, one top-level directory each, per the repo's commit convention.

### DIFF
--- a/catalog/apps/glance/Launchfile
+++ b/catalog/apps/glance/Launchfile
@@ -14,4 +14,5 @@ storage:
   config:
     path: /app/data
     persistent: true
+health: /api/healthz
 restart: always
--- a/catalog/apps/listmonk/Launchfile
+++ b/catalog/apps/listmonk/Launchfile
@@ -29,5 +29,10 @@ env:
     generator: secret
     sensitive: true
     description: "Admin password"
-health: /api/health
+  LISTMONK_db__ssl_mode:
+    default: "disable"
+    description: "Postgres SSL mode. listmonk sets no default and omits empty fields from the DSN, so without this the release command connects with sslmode unset"
+commands:
+  release: "./listmonk --install --idempotent --yes --config '' && ./listmonk --upgrade --yes --config ''"
+health: /health
 restart: always
--- a/catalog/apps/listmonk/metadata.yaml
+++ b/catalog/apps/listmonk/metadata.yaml
@@ -2,18 +2,18 @@ category: Email
 tagline: High-performance self-hosted newsletter and mailing list manager
 homepage: https://listmonk.app
 test_results:
-  last_tested: 2026-04-08
-  pull_time_seconds: 3
-  startup_time_seconds: 8
-  total_disk_mb: 292
-  health_check_passed: true
-  notes: ""
+  last_tested: 2026-08-28
+  pull_time_seconds: 2
+  startup_time_seconds: 7
+  total_disk_mb: 123
+  health_check_passed: false
+  notes: Harness result only. The harness does not execute commands.release, so listmonk starts over an uninstalled schema and never serves. Verified passing with the reference Docker provider, which does run release (D-48).
 images:
   - name: listmonk/listmonk:latest
-    size_mb: 33
+    size_mb: 14
     platform:
       - linux/arm64
   - name: postgres:16-alpine
-    size_mb: 259
+    size_mb: 109
     platform:
       - linux/arm64
--- a/catalog/apps/rallly/Launchfile
+++ b/catalog/apps/rallly/Launchfile
@@ -20,4 +20,5 @@ env:
     sensitive: true
   NEXT_PUBLIC_BASE_URL:
     default: "http://localhost:3000"
+health: /api/status
 restart: always
--- a/spec/SPEC.md
+++ b/spec/SPEC.md
@@ -451,6 +451,8 @@ Platforms execute well-known commands in this order: **build → release → sta
 | `seed` | Seed the database with initial data | On demand (first deploy or explicit trigger) |
 | `test` | Run the test suite | On demand (CI or explicit trigger) |
 
+One-time initialization that must complete before the app serves (schema install, first-run setup) belongs in `release`. Guard it with the app's own idempotent flags (install-if-needed) so it is safe to run on every deploy — the same idempotency recommendation the spec makes for [bootstrap](#bootstrap-stage). The stakes differ: `bootstrap` is user-invoked and non-deploy-failing, while `release` runs automatically on every deploy and a non-zero exit fails that deploy.
+
 Additional named commands are allowed and invoked on demand.
 
 ```yaml


ziadsawalha:
## Addressed — with a correction to the premise

All three blockers are fixed. Blocker 1 changed shape once I actually deployed it: the release does fail, but so does listmonk on `main` today.

### Blocker 1 — deployed. It failed. Then I fixed the entry.

`launchfile up catalog/apps/listmonk --docker`, reference Docker provider, Docker Desktop 29.7.2, `listmonk/listmonk:latest` v6.2.0, linux/arm64.

**1. The release as declared exits 1.**

```
init.go:331: connecting to db: listmonk-postgres:5432/listmonk
init.go:361: error connecting to DB: pq: SSL is not enabled on the server
release [default] failed with exit code 1 — deploy aborted
```

`--config ''` makes listmonk read config from the environment only. listmonk sets no default for `db.ssl_mode`, and builds its DSN from non-empty fields only (`cmd/init.go`, "Build Postgres DSN conditionally with non-empty fields"), so `sslmode` never reaches the connection string and the driver's own default applies. Upstream's own `docker-compose.yml` pairs that exact command with `LISTMONK_db__ssl_mode: disable`. The entry copied the command and not the variable. Added it.

**2. listmonk does not deploy on `main` either.** Same provider, `origin/main`'s entry, no `release`:

```
main.go:159: the database does not appear to be setup. Run --install.
```

repeating in a restart loop until the health gate times out at 120s. So this declaration is not converting an app that deploys today into one whose deploy fails. It is the missing piece.

**3. `health: /api/health` is wrong, and it is already on `main`.** It is registered on the auth-middleware group in `cmd/handlers.go` and answers `403 {"message":"invalid session"}` unauthenticated. The public route — commented upstream as "Public health API endpoint" — is `/health`. Changed it.

**With all three fixed:**

```
  ↓ Release [default] via docker compose run --rm listmonk
  ✓ Release [default] complete
  ✓ Health check passed
  listmonk is running at http://localhost:9000
```

`GET /health` → `200 {"data":true}`. `GET /admin/login` → `200`. Both containers report `healthy`.

**Idempotency — the claim the spec paragraph actually makes.** Second deploy over the installed database:

```
** first time (idempotent) installation **
install.go:53: skipping install as database appears to be already setup
main.go:165: running upgrade...
upgrade.go:74: no upgrades to run. Database is up to date.
```

Exit 0, health passed.

### Blocker 2 — rewritten, not appended

The "the reference docker provider does not yet execute release" sentence is gone from the history. The new body states that the Docker provider executes `release` as of D-48 (`providers/docker/src/release.ts`), that this declaration therefore changes listmonk's deploy, and that a non-zero exit fails it per `PROVIDERS.md` §10.10.

### Blocker 3 — suggestion applied verbatim

### On the two health checks you cleared

- **rallly** — verified live rather than from source: `200`, `{"status":"ok",...,"database":"connected"}`, container healthy. Merges as reviewed.
- **glance** — the path is right (`internal/glance/glance.go:458`), but I could not observe it, because glance never serves. The entry mounts its volume at `/app/data` while the image entrypoint is `--config /app/config/glance.yml`, so it crash-loops on `parsing config: reading /app/config/glance.yml: no such file or directory`. Pre-existing defect in the entry, untouched here; filed as #324. Say the word if you would rather I split the glance line out than ship a health path for an app that cannot reach it.

### Two findings outside this PR — filed as #324 and #325

**The catalog harness cannot validate any app that declares `release`.** `catalog/test/src/` contains no reference to `commands.release` — the harness does its own Launchfile translation and drops the stage. I ran `bun run src/test-app.ts listmonk`; it reproduced exactly the crash loop this PR fixes and wrote `health_check_passed: false`. That record is in the commit, with only the `notes` field hand-edited, to say why the `false` does not describe the shipped provider. The same gap means chatwoot and hedgedoc-v2 can never record a passing harness result either.

**The health gate passes a crash-looping container (#325).** Deploying glance printed `✓ Health check passed` while the container sat in `Restarting (1)`. Reproduced on every attempt. The cause is `isContainerHealthy` (`providers/docker/src/provider.ts:825`) treating an empty `Health` as "declares no healthcheck":

```js
const declaresNoCheck = container.Health === "";
return container.State === "running" && (container.Health === "healthy" || declaresNoCheck);
```

Sampling `docker compose ps --format json` every 150ms through a crash loop shows the container flapping through short `running` windows where `Health` is empty even though the generated compose *does* declare a healthcheck:

```
t=  1.64s  State='running'     Health=''
t=  1.87s  State='restarting'  Health=''
t=  2.16s  State='running'     Health=''
t=  2.39s  State='restarting'  Health=''
t=  5.25s  State='restarting'  Health='starting'
```

Any poll landing in one of those ~230ms windows returns `ok: true`. `Health === ""` is overloaded — it means both "no healthcheck declared" and "healthcheck declared, not yet evaluated" — and the provider already knows which services it gave a healthcheck to, so it should not be inferring that from the string. This is a mechanism by which `health_check_passed: true` can certify an app that never serves.

### State

Rebased onto `affbc13`. Two commits, one top-level directory each. Catalog tests 33/33.

---

*Edited after posting, twice. (1) The `ssl_mode` paragraph originally said listmonk's built-in default is `require`; it sets no default at all — it omits empty fields from the DSN, so the driver decides. Corrected in the commit body and the field description too (`749a030`). (2) The health-gate finding originally blamed `start_period`. That was wrong — I had not read the gate. Replaced with the actual cause, read from `provider.ts:825` and reproduced. The observed failures and the fixes are unchanged in both cases.*

